PR #1, opened 2020-10-21
Create info page

miyamoto-san:


.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+dist
```
kaj, line 2: It might be a good idea to change the editor config so the last line in each file is ended with a proper newline.

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "",
   "main": "app.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "style": "cp -r ./node_modules/kth-style/dist ./dist"
```
me97esn, line 8: I would prefer a more descriptive name of this test, like copy-style. It's easy to think it formats the code or something else with this short name

kaj (commented):
One minor nitpick, other than that this looks fine!

kaj (approved):
Looks good to me now (The conditional in the template is still broken, but there is a separate card about that later).

package.json
miyamoto-san: This is not a final solution so I don't know if it's worth it really but otherwise I think you're absolutely correct, that is a better name! :)

exacs: Just mentioning that the last commits change the name for `style` to `build:style` and add a `postinstall` script

PR #7, opened 2020-11-13
Implemented a session handler

miyamoto-san:
I've got a few todos that I want to look into before I call this a proper PR but until then please check out and comment on my session handling middleware

### DIFF
--- a/src/middleware/authenticationMiddleware.js
+++ b/src/middleware/authenticationMiddleware.js
@@ -1,0 +1,32 @@
+const got = require("got");
+const { validate, parse } = require("fast-xml-parser");
+
+async function authenticationMiddleware(req, _res, next) {
+  if (!req.session.userId) {
+    const baseUrl = `${process.env.SERVER_HOST_URL}/kpm`;
+    const serviceValidateUrl = new URL(
+      `${process.env.SSO_ROOT_URL}/serviceValidate`
+    );
+
+    if (req.query.ticket) {
+      serviceValidateUrl.searchParams.set("ticket", req.query.ticket);
+      serviceValidateUrl.searchParams.set("service", baseUrl);
+
+      const { body } = await got(serviceValidateUrl);
+      const parsedBody = parse(body);
+
+      if (
+        validate(body) === true &&
+        !parsedBody["cas:serviceResponse"]["cas:authenticationFailure"]
+      ) {
+        req.session.userId =
+          parsedBody["cas:serviceResponse"]["cas:authenticationSuccess"][
+            "cas:user"
+          ];
+      }
+    }
+  }
+  next();
+}
+
+module.exports = authenticationMiddleware;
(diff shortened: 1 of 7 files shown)


src/middleware/authenticationMiddleware.js
```diff
@@ -0,0 +1,32 @@
...
+        !parsedBody["cas:serviceResponse"]["cas:authenticationFailure"]
+      ) {
+        req.session.userId =
+          parsedBody["cas:serviceResponse"]["cas:authenticationSuccess"][
+            "cas:user"
+          ];
+      }
+    }
+  }
+  next();
```
miyamoto-san, line 29: I'm going to add in a try/catch block here as well to handle exceptions that may arise

src/server.js
```diff
@@ -61,8 +62,27 @@ async function fetchBlock(str) {
   return res.body;
 }
 
+app.set("trust proxy", 1); // TODO: note to self: what does this actually do?
```
miyamoto-san, line 65: https://expressjs.com/en/guide/behind-proxies.html

exacs: Rather than knowing **what** the trust proxy setup does, I'd like to know **why** is it relevant for our app

src/server.js
```diff
@@ -2,8 +2,8 @@ const handlebars = require("handlebars");
 const path = require("path");
 const fs = require("fs");
 const crypto = require("crypto");
+const session = require("cookie-session");
```
exacs, line 5: As I read, this library stores the session in the client. Haven't we say to store it in the server? Using memcached, redis or similar?

src/server.js
```diff
@@ -61,8 +62,27 @@ async function fetchBlock(str) {
   return res.body;
 }
 
+app.set("trust proxy", 1); // TODO: note to self: what does this actually do?
+
+var expires = new Date(Date.now() + 60 * 60 * 1000); // TODO: is 1h fine?
```
exacs, line 67: That's a good question but we don't need to answer it now. Let's leave the TODO for now :)

exacs (commented):
Overall this works but this doesn't work when we set the KPM from outside /kpm.

For example, if I click "login" from `canvas.kth.se`, this PR doesn't implement anything that takes me back to `canvas.kth.se`

src/server.js
kaj: At least a session key needs to be on the client.  If we can make the session small enough, and have a safe way of validating it on the server, I see no real problem with using a cookie as the main session storage.

src/server.js
miyamoto-san: Sure, but the why might be derived from the why. As far as I understand it allow us to be behind a proxy, as for example traefik, which seems like a good idea :)

src/server.js
miyamoto-san: > Although the app will not fail to run if the application variable trust proxy is not set, it will incorrectly register the proxy’s IP address as the client IP address unless trust proxy is configured.

src/server.js
miyamoto-san: > Although the app will not fail to run if the application variable trust proxy is not set, it will incorrectly register the proxy’s IP address as the client IP address unless trust proxy is configured.

PR #9, opened 2020-11-17
Redirect after login

exacs:


kaj (approved):
Ser rimligt ut, tror jag.

src/kpm-loggedout.js.handlebars
```diff
@@ -1,9 +1,10 @@
 (function(d){
   var kpm;
   function create_menu() {
+    var loginUrl = "{{loginUrl}}?next=" + window.encodeURI(window.location)
```
me97esn, line 4: This is a bit confusing, that there are two different variables with the same name, but in different contexts. One of them should probably be renamed.

me97esn: Actually, for me it was _very_ confusing at first :)

src/login-router.js
```diff
@@ -0,0 +1,72 @@
...
+  const loginUrl = new URL(`${process.env.SSO_ROOT_URL}/login`);
+  loginUrl.searchParams.set("service", serviceUrl);
+
+  log.info(`Redirect user to ${loginUrl}`);
+  res.redirect(loginUrl);
+});
+
+loginRouter.get("/callback", async (req, res) => {
+  try {
+    if (!req.session.userId) {
```
me97esn, line 25: Why is this check needed? Can you add a comment about it? And why don't we do anything if req.session.userId is set?

src/login-router.js
```diff
@@ -0,0 +1,72 @@
...
+        serviceValidateUrl.searchParams.set("ticket", req.query.ticket);
+        serviceValidateUrl.searchParams.set("service", serviceUrl);
+
+        const { body } = await got(serviceValidateUrl);
+        const parsedBody = parse(body);
+
+        if (
+          validate(body) === true &&
+          !parsedBody["cas:serviceResponse"]["cas:authenticationFailure"]
+        ) {
```
me97esn, line 47: There is no `else` clause here. Shouldn't we at least log something in case of authentication failures or if the body cannot be validated?

PR #11, opened 2020-11-25
Parcel and development tools

exacs:
This PR includes several changes.

### Parcel and Nodemon. Development improvements

- Adds [Parcel.js](https://parceljs.org/) to build client-side JS code.
- Adds a script `/client/kpm.js` to show how to include CSS files from a JS file using `import` and to see how Parcel changes the filename for the `menu2.css` file when creating the bundles
- Adds [Nodemon](https://github.com/remy/nodemon) to automatically restart the server when it changes
- Adds [Concurrently](https://www.npmjs.com/package/concurrently) which starts both nodemon and parcel in watch mode.
- Adds npm scripts: `dev`, `dev:client`, `dev:server`

### Changes in Docker related files

- Adds a build step in `Dockerfile`
- Creates a three-stage Dockerfile (development, builder and production)
- Now the `npm run build` command is not run when installing things. It needs to be called manually.

### Other changes

- Splits the code in `/src` in two directories: `/client` and `/server`
- Renames variables and filenames
- Fixes the `build:kth-style` script in package.json
- Adds npm script: `clean`


Dockerfile
```diff
@@ -1,18 +1,39 @@
...
 RUN npm ci --production --unsafe-perm
 
-# Everything else
+# Third "stage" is the production image, where we don't install dependencies
+# but use the already installed ones.
+#
+# This way we can deliver an image without the toolchain (python, make, etc)
+FROM kthse/kth-nodejs:12.0.0 AS production
+COPY --from=builder node_modules node_modules
+COPY --from=development /usr/src/app/dist dist
```
kaj, line 36: Hm.  Will any node modules that uses natively built code put their `libwhatever.so` etc files in node_modules?  Or does running as root and using `--unsafe-perm` imply that it might put stuff under `/usr/local` ?

app.js
```diff
@@ -7,7 +7,7 @@ require("skog/bunyan").createLogger({
   serializers: require("bunyan").stdSerializers,
 });
 
-const server = require("./src/server");
+const server = require("./server/server");
```
kaj, line 10: Does `app.js` need to be in the project root?  I think it would be cleaner to have it in the `server` directory?

client/kpm.js
```diff
@@ -0,0 +1,4 @@
+console.log("Hello from KPM.js");
+import("./menu2.css").then((r) => {
+  console.log("menu.css has been loaded!");
+});
```
kaj, line 4: This is just a demo for what we should do instead of using handlebars for kpm.js, right?  Then I think the demo should be dropped and the thing should be done for real before merging.  Also, why a demo-specific `menu2.css` when we have a `menu.css` ?

package.json
```diff
@@ -2,14 +2,19 @@
   "name": "kpm",
   "version": "1.0.0",
   "description": "",
-  "main": "app.js",
+  "//main": "app.js",
```
kaj, line 5: Is `"//main"` a way to comment out a line property?  Then maybe just remove it instead?  Or actually, take it back but use it to put `app.js` in the `server` directory?

server/info-page.handlebars
```diff
@@ -12,6 +12,7 @@
   <!--script src="/kpm/dist/js/menus.js"></script-->
   <!--script src="/kpm/dist/js/backtotop.js"></script-->
   <script src="/kpm/kpm.js" defer></script>
+  <script src="/kpm/dist/kpm.js" defer></script>
```
kaj, line 15: The `dist` directory is an implementation detail of parcel.  Do we really need to dirty our urls by exposing that to users?

server/server.js
```diff
@@ -47,12 +47,18 @@ function compileTemplate(name) {
...
-const template = compileTemplate("index.handlebars");
-const loggedInTemplate = compileTemplate("kpm-loggedin.js.handlebars");
-const loggedOutTemplate = compileTemplate("kpm-loggedout.js.handlebars");
+const infoPageTemplate = compileTemplate("info-page.handlebars");
+const loggedInTemplate = compileTemplate(
+  "../client/kpm-loggedin.js.handlebars"
+);
+const loggedOutTemplate = compileTemplate(
+  "../client/kpm-loggedout.js.handlebars"
+);
```
kaj, line 56: These should be removed before this PR is done, right?

server/server.js
```diff
@@ -47,12 +47,18 @@ function compileTemplate(name) {
...
 
 // TODO: Sass?
-const menuCssData = fs.readFileSync(path.resolve(__dirname, "menu.css"));
+const menuCssData = fs.readFileSync(
+  path.resolve(__dirname, "../client/menu.css")
+);
```
kaj, line 61: And this should also not be needed anymore?

kaj (commented):
I have a bunch of questions ...

client/kpm.js
exacs: Yes, this is just for demo. It is written to see how Parcel:

- converts `import(...css)` into a fetch, `<link>` tag, etc. 
- to see how it transpiles `menu2.css` into something called `menu2.[hash].css`

I thogut this PR is big enough and didn't want to change the logic in `kpm.js` with handlebars yet because it might include slight design changes.

app.js
exacs: That's a good discussion. Should the `app` be the same as the `server` or should be splitted? In Node.js I think is more common to have it separately so that `app.js` can initialize things that are unrelated to the server like the logger, `dotenv`, etc.

Dockerfile
exacs: That's a good question and I don't have the answer now

kaj: I _think_ any libraries compiled by npm will end up in `node_modules`, so it should be enough.  If we end upp adding any run-time dependencies with apk, we just have to remember to add them in all the stages.  So this should be ok.

package.json
exacs: Ouch, this is something I left mistakenly. Parcel 2 transpiles JS and uses the `main` field in package.json as file destination, so I commented out because we don't want to use that `main` field like that.

Then I've just downgraded to Parcel 1 because Parcel 2 was still in beta and didn't have the problem anymore. So, I'll say we can keep as it was before.

server/info-page.handlebars
exacs: This is for the demo `js` file. Sure `/kpm/kpm.js` is better than `/kpm/dist/kpm.js` :)

app.js
me97esn: I think it's best to keep it in the project root, only because that's the structure of all of our other (lms) apps

client/kpm.js
kaj: The concern about the size of the PR is valid.   But ideally I think each PR should put the project in a state where it is at least as releasable as it was before, and I don't think we should release the pink background ...  :sunglasses:

me97esn: What, Parcel creates `fetch` and link tags? That sounds really nice (also a bit too magical and scary...) actually. I have never been a fan of the "include css in js files" that has become the standard these last years.

exacs: > The concern about the size of the PR is valid. But ideally I think each PR should put the project in a state where it is at least as releasable as it was before, and I don't think we should release the pink background ... 😎

Ouch! Yes, you are right. I can change the `menu2.css` to have a more subtle change

exacs: > What, Parcel creates `fetch` and link tags? That sounds really nice (also a bit too magical and scary...) actually. I have never been a fan of the "include css in js files" that has become the standard these last years.

@me97esn yes, it fetches and adds the link tags automatically. You can inspect and see that `menu2.[hash].css` is fetched

app.js
```diff
@@ -7,7 +7,7 @@ require("skog/bunyan").createLogger({
...
 const log = require("skog");
 server.listen(process.env.PORT || 3000, () => {
   log.info("Started HTTP server");
```
kaj, line 13: One more question, after actually running the code:  Can we add the server url to this line?  Either just

```js
const port = process.env.PORT || 3000
server.listen(port, () => {
  log.info(`Started server on http://localhost:${port}/kpm/`);
});
```

... or if we can get `server` to report its actual url in a nice way?

miyamoto-san (approved):

PR #12, opened 2020-12-03
JS without handlebars

exacs:
This PR creates [`kpm.js` without handlebars](https://github.com/KTH/kpm/blob/modern-js/client/kpm.js) 🎉 

Moves lots of HTML to the backend under routes called `/panels`. See [the `server/panels` directory](https://github.com/KTH/kpm/tree/modern-js/server/panels). Then `kpm.js` will fetch those files and display them.

Small changes:

- `/dist` directory is mounted in `/`. So `/kpm/dist/something.css` → `/kpm/something.css`
- KTH Style related files are moved to its own directory. `/kpm/dist/js/vendor.js` → `kpm/kth-style/js/vendor.js`


server/panels/index-loggedin.handlebars
```diff
@@ -0,0 +1,14 @@
+<div class="kpmbar">
+  <a class="kpm-open-menu" href="#kpm">Menu</a>
+</div>
+<div class="kpmpanel"></div>
+{{#if message}}
+<div class="kpm-alert">
+  <div class="container">
+    <div>{{message}}</div>
```
kaj, line 8: Här var det väl `{{{message}}}` i motsvarande gammal kod, dvs med möjlighet till markup i själva meddelandet?  Jag hade i alla fall ett meddelande med en länk i min gamla `.env`, och det fungerar inte nu.

server/server.js
```diff
@@ -39,33 +38,11 @@ const blocks = {
...
-  "../client/kpm-loggedout.js.handlebars"
-);
-
-// TODO: Sass?
-const menuCssData = fs.readFileSync(
-  path.resolve(__dirname, "../client/menu.css")
+const infoPageTemplate = handlebars.compile(
+  fs.readFileSync(path.resolve(__dirname, "info-page.handlebars"), {
+    encoding: "utf-8",
+  })
```
kaj, line 44: Hmm ...  This _looks_ duplicate from function `compilePanel` in `panels/router.js`, but isn't really, since `__dirname` is different there.  Maybe we should have a function `compileTemplate` that takes a resolved path as its argument?

kaj (commented):
:+1: I like how the actual `kpm.js` is now the same if logged in or not, and `server/panels/router.js` is starting to look like it could do real work.  Good job!

Got a few quesions / comments for specifik paces in the code:

server/panels/index-loggedin.handlebars
miyamoto-san: @kaj jag uppdaterar detta till `{{{message}}}` så att HTML markup fungerar i det!

client/kpm.js
```diff
@@ -1,4 +1,54 @@
...
+  event.preventDefault();
+  event.stopPropagation();
+  kpm.querySelector(".kpmpanel").innerHTML = "<p>Loading...</p>";
+  kpm.querySelector(".kpmpanel").innerHTML = await fetchPanel("hello");
+}
+
+async function fetchPanel(panel) {
+  try {
+    const content = await window
+      .fetch(`/kpm/panels/${panel}`)
```
kaj, line 39: Den här kommer väl att utgå från "sidans" basurl och inte scriptets?  Eller gör parcel någon magi med det?

client/kpm.js
```diff
@@ -36,7 +36,7 @@ async function openMenu(event) {
 async function fetchPanel(panel) {
   try {
     const content = await window
-      .fetch(`/kpm/panels/${panel}`)
+      .fetch(`${process.env.SERVER_HOST_URL}/kpm/panels/${panel}`)
```
kaj: Wait what?  This is client code, so I don't think that variable is available?

kaj: Or does parcel apply magic here?  Does it now which variables to do magic with?  In the code my client sees, it says:

```js
            return window.fetch("".concat("http://localdev.kth.se:3000", "/kpm/panels/").concat(panel)).then(function (r) {
```

So ...  I guess it seems to work, even though it's a bit more "magic" than I would like ..?  How do I know which variables are evaluated at runtime (in the client) and which are evaluated at build time (in the server)?

miyamoto-san: So, basically anything that is compiled by Parcel (i.e. client code) will have environment variables evaluated in compile time. There will always be a compilation of that code before it hits the end user and as such all environment variables will always have been evaluated!

miyamoto-san: The thing is, this code basically isn't _really_ client code. The client code is what will be compiled and sent to them. They never directly interact with this code but only with a _compiled version_ of this code!

server/panels/router.js
```diff
@@ -0,0 +1,46 @@
...
+    res.send(
+      indexLoggedOut({
+        loginUrl: `${process.env.SERVER_HOST_URL}/kpm/login`,
+        message: process.env.LOGGED_OUT_ALERT,
+      })
+    );
+  }
+});
+
+panelsRouter.get("/hello", (req, res) => {
```
me97esn, line 34: Should this route still exist? Or should we delete it?

me97esn:
Got these errors in devtools in FF:
`The resource from “http://localdev.kth.se:3000/kpm/kth-style/css/kth-bootstrap.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).` 

`The resource from “http://localdev.kth.se:3000/kpm/kth-style/js/vendor.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).`

Guess we should fix them, set the correct content-type?

me97esn:
Is this warning something we should fix:
`Cookie “kpm” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite`?

client/kpm.js
kaj: Yes, I'm avare of that.  :-)  The question is how parcel knows which variables to evaluate at compile time and which to keep for run-time.

kaj: Also, I find it rather strange that after eavluating `process.env.SERVER_HOST_URL` to a static value, it still evaluates ` ${process.env.SERVER_HOST_URL}/kpm/panels/` to `"".concat("static value", "static part")`.  Why not do that concat compile-time as well?

server/panels/router.js
kaj: This is the route that serves the "hello" menu panel, so should be kept at least until we implement some more useful menu panel(s).

kaj:
> Got these errors in devtools in FF:
> `The resource from “http://localdev.kth.se:3000/kpm/kth-style/css/kth-bootstrap.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).`
> 
> `The resource from “http://localdev.kth.se:3000/kpm/kth-style/js/vendor.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).`

So we send some of our css and javascript with a text/html content-type?  Yes, that should be fixed.



> Is this warning something we should fix:
> `Cookie “kpm” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite`?

Yes, we should set the "secure" attribute.  But not when running http on localhost, so if you only get that when testing locally I think it might be ok.

client/kpm.js
```diff
@@ -1,4 +1,54 @@
...
+  try {
+    const content = await window
+      .fetch(`${process.env.SERVER_HOST_URL}/kpm/panels/${panel}`)
+      .then((r) => {
+        if (r.status > 400) {
+          throw new Error(`${r.status} ${r.statusText}`);
+        }
+        return r;
+      })
+      .then((r) => r.text());
```
me97esn, line 46: I think we shouldn't mix async/await with promise chains. Use either or. And since we use async/await all over the place, we should use it here as well.

me97esn: I also don't like the error handling here. We manually throw an error, and then we catch the same error in the same function and just logs it to the console. Why throw it if we only want to log it anyways? It would make sense to throw an error if we were to catch it in another function higher up in the call chain, but that's not the case here.

me97esn: And another question about the error handling: we do have an error.handlebars panel template. When should that be used? Here we console.error when we get an error, but since we have an error panel I believe that we sometimes show the errors to the end user. When do we console.error and when do we render errors?

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "plugins": ["@babel/plugin-transform-runtime"]
```
me97esn, line 2: I looked at the docs for this plugin:
`Why?

Babel uses very small helpers for common functions such as _extend. By default this will be added to every file that requires it. This duplication is sometimes unnecessary, especially when your application is spread out over multiple files.

This is where the @babel/plugin-transform-runtime plugin comes in: all of the helpers will reference the module @babel/runtime to avoid duplication across your compiled output. The runtime will be compiled into your build.`

Since we compile the menu to one file (kpm.js), why is this plugin needed?

client/kpm.js
```diff
@@ -1,4 +1,54 @@
...
+
+    return content;
+  } catch (err) {
+    console.error(`Error when fetching the "${panel}" panel: `, err);
```
me97esn, line 50: Question about the error handling: we do have an error.handlebars panel template. When should that be used? Here we console.error when we get an error, but since we have an error panel I believe that we sometimes show the errors to the end user. When do we console.error and when do we render errors?

server/panels/error.handlebars
```diff
@@ -0,0 +1 @@
+<p>Error</p>
```
me97esn, line 1: Let's have a discussion when this template should be used, and how. If we throw an error in the server side, will that cause this panel to render? We currently send this panel with statuscode 200, is that correct? Shouldn't it be send with statuscode 500?

server/panels/router.js
```diff
@@ -0,0 +1,46 @@
...
+  if (req.session.userId) {
+    res.send(
+      helloPanel({
+        userName: req.session.userId,
+        infoUrl: `${process.env.SERVER_HOST_URL}/kpm/`,
+        logoutUrl: `${process.env.SERVER_HOST_URL}/kpm/logout`,
+      })
+    );
+  } else {
+    res.send(errorPanel());
```
me97esn, line 44: Is this how we want to handle errors? Should we spread the logic of how to handle errors all over the codebase (even if it's only one line, it's still duplicates when we spread it to other places. If we decide to also send status:500 we will have to change it everywhere we handle errors in the server)? Or should we have a top level handling of the errors, and thats where we render the errorPanel? I prefer the latter, but I want to hear your opinions here.

me97esn (changes requested):
I would at least have a discussion about my comments. If we keep it as it is, I want to know the reasons.

client/kpm.js
kaj: I kind of agree on the error handling.  What does this fetch function do in case of network time out etc?  If it can raise an exception we still need to catch it (in the existing personal menu, there is callbacks on XMLTttpRequest in [a sendrequest function](https://gita.sys.kth.se/Infosys/social/blob/834c998135e2a96ada1363b087df1dc797a55579/social/templates/toolbar/widget.js#L294-L328) which takes callbacks for success or failure [when called](https://gita.sys.kth.se/Infosys/social/blob/834c998135e2a96ada1363b087df1dc797a55579/social/templates/toolbar/widget.js#L189-L207)).

When you say "mix async/await with promise chains", do you refer to the fact there there is both `wait` and `.then(...)` in the code?  That feels quite natural to me (but then Rust makes it feel more natural by using a postfix dot notation for the `await`).

me97esn:
I also just realized that the x to close the info message doesn't work anymore. That should be fixed.

.babelrc
kaj: Agreed.  Also, looking at that file, there seems to be a lot of (generated) code related to module loading, which I hope isn't really used.  If that can be made shorter, that would be good.

miyamoto-san: I doubt that we won't split the file up eventually on the other hand!

client/kpm.js
kaj: When an error occurs on the server side, the error template may be used, but the response status will also be >= 400.  So the _content_ of the error response will probably not be exposed (other than to developers replaying the request by hand).  There should be a error message shown to the user, but that needs to be generated by the client-side javascript (since the error might be a network error, so we may not have a response at all).  [Here's how the old code does that](https://gita.sys.kth.se/Infosys/social/blob/834c998135e2a96ada1363b087df1dc797a55579/social/templates/toolbar/widget.js#L282-L285).

server/panels/error.handlebars
kaj: Sending an error with status code 200 is never correct!  Good catch! 😎

me97esn:
Another question. Perhaps not necessary to fix right now, but would be really nice: Why don't we have source maps in the browser? When looking at the compiled file /dist/kpm.js, it looks like it is prepared for a sourcemap file. But the browser doesn't download it, hence debugging of the uncompiled code doesn't work.
And: since it looks like sourcemaps are somewhat enabled, are they also enabled in production code? Or only in develepment code?

client/kpm.js
```diff
@@ -39,8 +39,9 @@ async function fetchPanel(panel) {
   );
   if (response.status > 400) {
     console.error(`Error when fetching the "${panel}" panel: `, err);
+  } else {
+    return await response.text();
   }
-  return await response.text();
```
kaj: Wait what?  Now if there is an error, this function just returns undefined?  I think it either needs to tell its caller there was an error or return an error message.  Since the caller just does `element.innerHtml = await fetchPanel(name)` it should probably return a short html-formatted error message.

Note that if the same function is used to load the top-level menu bar, the error message needs to be a one-liner when rendered.

me97esn: Yes, that's a good comment. Let's plan a meeting next week to discuss how to handle errors in this app, both client side and server side. Should we always return a html text to display to the user? We should all agree on a structured way to handle these errors. 
I have actually not changed _how_ the error is handled since before. It's just more obvious now that we don't return anything if there's an error.

.babelrc
kaj: Spitting the source of that file is one thing that may be ok (if it getts larger enough that it is relevant).  But I think we should think twice before having the client load multiple js files instead of one (unless it's one file for all browsers plus one file of polyfills that only antique browsers needs to load (the old personal menu does that)).

me97esn: I agree with Rasmus here. Splitting the file would probably effect the performance negativily, since every browser have a limit on the concurrent requests it can perform (usually 5 if I'm not mistaken). So if we would have 5 different js files, it would cause lots of other file to wait until the js files for the menu is loaded.

miyamoto-san: I believe you're correct @me97esn but I don't think that's how we should load the JS in case we're to split up the file, not saying we are but, if we were to find that necessary in the future we'd load the JS programmatically when needed. I'll attach a code example of what I mean.

```js
document.addEventListener("DOMContentLoaded", () => {
  const head = document.querySelector('head');
  const load = (src) => {
    const script = document.createElement('script');
    script.src = src;
    head.appendChild(script);
  }
  if (some condition) {
    load('/myAwesomeJS.js')
  }
});
```
I think it makes sense to just load functionality when it is needed. For example most of the JS file won't ever be executed if you don't login. Why should we then send that to the user? I suggest we only send it when needed.

me97esn:
Question about the content-type errors: they seem to come from the file info-page.handlebars (at least some of them). What's the deal with this page? Should this page include kth-style? I guess it should, right? The links are broken at the moment, do we need to fix them?

.babelrc
kaj: I think the complexity of the above example is worth about 1 kB (maybe slightly less), and the round-trip for each extra request is worth about 5 kB (maybe a log more).  So if the saving in the allways-loaded js less than 6 kB (plus whatever cost is introduced by `@babel/plugin-transform-runtime` itself and similar tooling) it's not worth it.  :sunglasses:

miyamoto-san: Well assuming we're to build the menu exactly like the old one is built that is. I don't think it's necessary to keep the JS to a strict 14kb limit :)

me97esn:
Ok, the content-type errors (which actually was 404 errors) are now fixed. There's still one thing left that should be fixed in this PR. The 'x' to close the message worked before this PR, nothing happens when we click it now. Perhaps @miyamoto-san can take a look at it next week? Since everything that worked before a PR should work after the PR, this should be fixed before we merge.

me97esn (changes requested):
The 'x' to close the message worked before this PR, nothing happens when we click it now. Perhaps @miyamoto-san can take a look at it next week? Since everything that worked before a PR should work after the PR, this should be fixed before we merge.

me97esn:
I found one thing about this Parcel code that bothers me a bit:
We build our code with one version of Parcel, defined in our package.json.
Then we include kth-style, or more precisely the transpiled kth-style file in `node_modules/kth-style/dist/js/vendor.js`. This file is also build with Parcel, but with the version defined in `kth-style/package.json`, currently `"parcel-bundler": "^1.12.4"`.
It works now, but I can foresee really weird errors occuring if we combine code build with different parcel versions. I think we should transpire the vendor.js file ourselves instead of using the pre-transpiled file. That would also allow us to debug the file if an error occurs somewhere in this file (as actually did today, that's why I realized this problem), by adding a source map file to the vendor.js file. As it is now, it's virtually impossible to debug.

server/info-page.handlebars
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+!DOCTYPE html>
```
kaj, line 1: Det här måste ju vara en typo.

me97esn: Haha hoppsan, det är det nog ja :) Ursäkta!

kaj:
> We build our code with one version of Parcel, defined in our package.json.
> Then we include kth-style, or more precisely the transpiled kth-style file in `node_modules/kth-style/dist/js/vendor.js`. This file is also build with Parcel, but with the version defined in `kth-style/package.json`

Kth-style and `vendor.js` is only used for the info page, and not for the personal menu itself.  As the personal menu should be usable on any page, including pages that may have a several years old version of kth-style, I don't think this is a problem, or if it seems to be a problem, there may be a deeper problem with using transpiled code for the personal menu at all?

me97esn:
> > We build our code with one version of Parcel, defined in our package.json.
> > Then we include kth-style, or more precisely the transpiled kth-style file in `node_modules/kth-style/dist/js/vendor.js`. This file is also build with Parcel, but with the version defined in `kth-style/package.json`
> 
> Kth-style and `vendor.js` is only used for the info page, and not for the personal menu itself. As the personal menu should be usable on any page, including pages that may have a several years old version of kth-style, I don't think this is a problem, or if it seems to be a problem, there may be a deeper problem with using transpiled code for the personal menu at all?

Yes, perhaps this is a bigger problem, or perhaps it's fine. But when looking at the transpiled code, it definitely looks like Parcel shares code between different bundles on the same page. So if kpm.js is included into places for instance, we will mix different Parcel versions. Could perhaps be problematic. Look at the code snippet from the transpiled code:
`// anything defined in a previous bundle is accessed via the
// orig method which is the require for previous bundles
parcelRequire = (function (modules, cache, entry, globalName) {
  // Save the require from previous bundle to this closure if any
`

client/kpm.js
exacs: I think the question is: "what should we do if the request fails?" Should we show some message to the user? Should we retry?

.babelrc
exacs: afaik `@babel/plugin-transform-runtime` includes polyfills and transformations for things like `async/await`, generators, etc. See: https://babeljs.io/docs/en/babel-plugin-transform-runtime

Btw, there is a technique to serve two different JS files: one for new browsers (small and with new syntax) and one for old browsers. Take a look:

```
<script type="module" src="new.js"></script>
<script nomodule src="old.js"></script>
```

me97esn:
FYI: we had a meeting about this PR today, came to the conclusion that we _couldn't_ come to a conclusion at the moment. I have scheduled another meeting on friday, where we can discuss a little bit more and hopefully come up with a plan forward.

PR #13, opened 2020-12-16
Internationalization

miyamoto-san:
This is my proposal for how we should handle internationalization in KPM on the client side. It's basically a function that returns and object with the translations.

```js
const translations = {
  en: {
    loading: "Loading",
    menu: "Menu",
    login: "Log in",
  },
  sv: {
    loading: "Laddar",
    menu: "Meny",
    login: "Logga in",
  },
};

/**
 *
 * @param {string} input input you want to translate
 * @param {string} lang language you want to translate into
 * @returns {string}
 */

function intl(input, lang) {
  return translations[lang][input];
}

module.exports = intl;
```
The idea is to connect this to a cookie so that the lang param is always whatever you've chosen and set as your language by means of cookie. The next iteration would probably be replacing lang with a request to cookies. 

I think this works for frontend translations but for the one were we've pre-generated the templates all I can think of is that we use either a url param or some sort of a header to let the server know which template we want. We can either compile an english template or use a similar function and simply send the strings dynamically from the server so to say...

### DIFF
--- a/client/kpm.js
+++ b/client/kpm.js
@@ -2,12 +2,12 @@ import { intl, addLanguageSelector } from "./translation";
 
 const kpm = document.createElement("nav");
 kpm.id = "kpm";
-kpm.innerHTML = `<div class="kpmbar">${intl("loading", "sv")}...</div>`;
+kpm.innerHTML = `<div class="kpmbar">${intl("loading")}...</div>`;
 document.body.insertBefore(kpm, document.body.firstChild);
 
 function recreate() {
   document.getElementById("kpm").childNodes.forEach((node) => node.remove());
-  kpm.innerHTML = `<div class="kpmbar">${intl("loading", "sv")}...</div>`;
+  kpm.innerHTML = `<div class="kpmbar">${intl("loading")}...</div>`;
   create();
 }
 
--- a/client/kpm.js
+++ b/client/kpm.js
@@ -1,13 +1,17 @@
-const intl = require("./translation");
+import { intl, addLanguageSelector } from "./translation";
 
 const kpm = document.createElement("nav");
 kpm.id = "kpm";
 kpm.innerHTML = `<div class="kpmbar">${intl("loading", "sv")}...</div>`;
 document.body.insertBefore(kpm, document.body.firstChild);
 
-function setLanguage() {}
+function recreate() {
+  document.getElementById("kpm").childNodes.forEach((node) => node.remove());
+  kpm.innerHTML = `<div class="kpmbar">${intl("loading", "sv")}...</div>`;
+  create();
+}
 
-async function start() {
+async function create() {
   await import("./menu.css");
   const content = await fetchPanel("");
   kpm.innerHTML = content;
@@ -16,7 +20,7 @@ async function start() {
   const openMenuButton = kpm.getElementsByClassName("kpm-open-menu")[0];
 
   if (loginButton) {
-    loginButton.textContent = intl("login", "sv");
+    loginButton.textContent = intl("login");
     const loginUrl = loginButton.getAttribute("href");
 
     loginButton.setAttribute(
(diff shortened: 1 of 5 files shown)


kaj:
This is close to what the old personal menu do at client-side, and I think it's fine for a few messages like that (the old personal menu has 11 messages of this kind, which is about as many as I think is ok before swithching to something more managed.

For the main content of the menu bar and panels (which I think should be localized at server side), I think we need a solution with more tooling for maintenence, like gettext or [fluent](https://www.projectfluent.org).

kaj:
In the old personal menu, the graph is directed the other way:
```javascript
        msglang = {
            'pm': { 'sv': 'Personliga menyn', 'en': 'Personal Menu' },
            'admc': { 'sv': 'Administrera kurs', 'en': 'Administrate course' },
            'addfav': { 'sv': "Lägg till favorit", 'en': "Add favorite" },
            // ...
        };
```

and the function to get a translation do not take the language, it knows it:

```javascript
    function msg(ref) {
        var m = msglang[ref];
        return m[KthToolbarConfig.language] || m['en'];
    }
```

```javascript
    pm.setAttribute('aria-label', msg('pm'));
```

Other than that, this looks rather similar to how we wrote it back then ...  :sunglasses:

client/menu.css
```diff
@@ -110,4 +110,4 @@ div#main.ic-Layout-columns > span > span,
   100% {
     height: auto;
   }
-};
+} ;
```
kaj: The space isn't the problem here, it just (aparently) makes the linter ignore the semicolon it should warn about.  :-)

miyamoto-san:
> In the old personal menu, the graph is directed the other way:
> 
> ```js
>         msglang = {
>             'pm': { 'sv': 'Personliga menyn', 'en': 'Personal Menu' },
>             'admc': { 'sv': 'Administrera kurs', 'en': 'Administrate course' },
>             'addfav': { 'sv': "Lägg till favorit", 'en': "Add favorite" },
>             // ...
>         };
> ```
> 
> and the function to get a translation do not take the language, it knows it:
> 
> ```js
>     function msg(ref) {
>         var m = msglang[ref];
>         return m[KthToolbarConfig.language] || m['en'];
>     }
> ```
> 
> ```js
>     pm.setAttribute('aria-label', msg('pm'));
> ```
> 
> Other than that, this looks rather similar to how we wrote it back then ... 😎

Yeah, more or less intentionally made it so :D

client/menu.css
miyamoto-san: I've never seen that happen before, I don't mind it though :))

client/kpm.js
```diff
@@ -1,9 +1,17 @@
+import { intl, addLanguageSelector } from "./translation";
+
 const kpm = document.createElement("nav");
 kpm.id = "kpm";
-kpm.innerHTML = '<div class="kpmbar">Loading...</div>';
+kpm.innerHTML = `<div class="kpmbar">${intl("loading", "sv")}...</div>`;
```
me97esn, line 5: Is this a typo or left from an earlier draft?
The function accepts one parameter, but this call sends two

client/kpm.js
```diff
@@ -1,9 +1,17 @@
...
 document.body.insertBefore(kpm, document.body.firstChild);
 
-async function start() {
+function recreate() {
+  document.getElementById("kpm").childNodes.forEach((node) => node.remove());
+  kpm.innerHTML = `<div class="kpmbar">${intl("loading", "sv")}...</div>`;
```
me97esn, line 10: Is this a typo or left from an earlier draft?
The function accepts one parameter, but this call sends two

client/translation.js
```diff
@@ -0,0 +1,51 @@
...
+    setLang("en");
+  } else {
+    setLang("sv");
+  }
+  document.getElementById("kpm-lang-selector").textContent = getLang();
+  recreate();
+}
+
+export function intl(input) {
+  const lang = Cookies.get("kpm_lang") || "sv";
```
me97esn, line 38: Is it correct that swedish should be the default language? Perhaps @kaj has an opinion about this?

client/kpm.js
miyamoto-san: Fixed this now! This wouldn't have been an issue had we used Typescript btw! But you're right, it was something from pre-xmas implementation that was left behind =)

client/translation.js
miyamoto-san: I have no idea but as I wrote this it made sense to me at least :)

miyamoto-san:
I've added a rebuild of the menu when you choose language. This is just to rerender some of the component with the new cookie set. I don't expect this to be our long term solution but as I discussed with @me97esn I think this is a reasonable approach as of now considering time and resources!

me97esn:
The code looks good IMO. I haven't tested it yet though, at least someone should do before we merge/deploy

exacs:
I've tested it and works! Don't see the need of deeper review so I'll merge it. 👍

PR #14, opened 2021-01-19
Make SERVER_HOST_URL known to kmp.js.

kaj:
The file `kpm.js` needs `SERVER_HOST_URL` to be known when processing it with parcel, so move the parcel processing from a docker build step to application startup.  Of course, this makes application startup slower, which may be a problem.

Babel and parcel packages moved from devDependencies to dependencies, to make them available in the "production" docker image.

The util-linux package in production image is required to get parcel running.

Temporarily disabled js-cookie in translation.js, as I failed to get parsel to resolv that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,23 @@
 # This Dockerfile uses multi-stage builds as recommended in
 # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md
 #
+# First "stage" is a development image, only used for unit testing
+FROM kthse/kth-nodejs:12.0.0 AS development
+WORKDIR /usr/src/app
+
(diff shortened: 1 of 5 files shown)


kaj:
An alternative solution might be to set `SERVER_HOST_URL` to some symbol like `'XXHOSTXX'` during build, and then just `sed` that to the real value at application startup.  But postprocessing generated files feels somewhat icky and would probably mess up the sourcemaps.

kaj:
Oh.  The unit tests used the development docker image (which I removed).  So should I get that back just for the tests?  Might as well.

kaj:
Since #15 is merged, I close this now.

PR #15, opened 2021-01-20
Don't use env in kpm.js

kaj:
Get the kpm base url from the relevant `script` tag instead of trying to evaluate a runtime environment variable in parcel.

There is still a problem with importing `menu.css`.  I have arranged things so that the menu works, but there is still an ugly stack trace in the console.   There is also a bunch of `consle.log("kpm: ...")` added. Most of them (if not all) should probably be removed later, but for now, until we get our fist 500 or so users, there's no harm in keeping them.

### DIFF
--- a/client/kpm.js
+++ b/client/kpm.js
@@ -1,9 +1,15 @@
+console.log("Starting kpm.js");
+
 import { intl, addLanguageSelector } from "./translation";
+
+console.log("Imported translation");
 
 // Note: src is the resolved url, not the raw attribute
 const scriptUrl = (
   document.currentScript || document.querySelector("script[src$='/kpm.js']")
 ).src;
+
+console.log("Script url is", scriptUrl);
 
 const kpm = document.createElement("nav");
 kpm.id = "kpm";
@@ -62,7 +68,9 @@ async function openMenu(event) {
 }
 
 async function fetchPanel(panel) {
-  const response = await window.fetch(new URL(`panels/${panel}`, scriptUrl));
+  url = new URL(`panels/${panel}`, scriptUrl);
+  console.log("Fetch panel", panel, "from", url);
+  const response = await window.fetch(url);
   if (response.status > 400) {
     console.error(`Error when fetching the "${panel}" panel: `, response);
   } else {
--- a/client/kpm.js
+++ b/client/kpm.js
@@ -51,9 +51,12 @@ async function create() {
   addLanguageSelector(recreate);
 
   const btn = document.getElementById("kpm-alert-btn");
-  btn.addEventListener("click", (e) => {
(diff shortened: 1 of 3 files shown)


kaj:
If we decide that this is good, #14 should be closed without merging.

exacs:
The code looks ok in my opinion 👍 . I'll write down a list of things to review before merging (I don't have the time today but I don't want to forget about it)

- [ ] Check the compatibility of [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and [Document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) 

     With that in mind, decide if maintain or remove `|| document.querySelector("script[src$='/kpm.js']")` in [Line 9](https://github.com/KTH/kpm/pull/15/files#diff-ee63f41c869919225ed3c71784dd1a6bc6bae9a87f7af43c475b2c9c4da1dc9eR9)

- [ ] Test the code locally and try the logs

miyamoto-san (approved):

exacs:
The code works locally. I wonder what happens if we do ```fetch(`panels/${panel}`)``` without passing the `scriptUrl`. Does `fetch` build URLs relative from the document or from the script?

I want to try this by embedding the script in Canvas without `scriptUrl` and see what happens

kaj:
> The code works locally. I wonder what happens if we do `` fetch(`panels/${panel}`) `` without passing the `scriptUrl`. Does `fetch` build URLs relative from the document or from the script?

From the document, so that wont work.

exacs:
`scriptUrl` doesn't seem to work now

For example, when accessing kpm from https://vc-dev-8-r.referens.sys.kth.se/social/:

```
kpm: Script url is https://vc-dev-8-r.referens.sys.kth.se/social/toolbar/widget.js
```

so the script is trying to fetch `https://vc-dev-8-r.referens.sys.kth.se/social/toolbar/panels/`

kaj:
> `kpm: Script url is https://vc-dev-8-r.referens.sys.kth.se/social/toolbar/widget.js`

Yes, but that is because social still uses a HTTP redirect when the cookie is set.  This is to be changed (mainly for cache-friendlyness resaons) in social, and then the script url in kpm will be correkt.

kaj:
Now there is code on https://vc-dev-8-r.referens.sys.kth.se/social/ to respect the cookie in the right way (clientside, without a redirect), and `scriptUrl` gets the correct value.

It still doesn't really work though, since kpm doesn't yet set any CORS headers. Should I add that in this PR, or should we merge this and make a separate PR for CORS headers?

kaj:
Så!  Efter att även ha lagt till både cors-headrar och options till fetch-anropet så verkar det funka för mig i alla fall!

PR #16, opened 2021-01-26
Implement a in-memory cache for Cortina blocks

exacs:
<details>
<summary>Benchmarks</summary>

Test: 1000 connections (100 concurrent) towards KPM deployed locally

## Without cache:

```
$ ab -n 1000 -c 100 http://localdev.kth.se:3000/kpm
  50%   4753
  66%   6217
  75%  21815
  80%  23063
  90%  24254
  95%  30507
  98%  84960
  99%  87397
 100%  100676 (longest request)
```

## With cache

```
$ ab -n 1000 -c 100 http://localdev.kth.se:3000/kpm
  50%    119
  66%    130
  75%    134
  80%    138
  90%    150
  95%    156
  98%    173
  99%    174
 100%    174 (longest request)
```
</details>

### DIFF
--- a/server/utils.js
+++ b/server/utils.js
@@ -1,6 +1,8 @@
 const handlebars = require("handlebars");
 const path = require("path");
 const fs = require("fs");
+const log = require("skog");
+const got = require("got");
 
 function compileTemplate(resolvePath, name) {
   return handlebars.compile(
@@ -10,4 +12,53 @@ function compileTemplate(resolvePath, name) {
   );
 }
 
-module.exports = { compileTemplate };
+// In-memory cached Cortina blocks. It stores Promises of blocks, not the
+// actual content
+const cachedBlocks = {};
+const blockIds = {
+  title: "1.260060",
+  megaMenu: "1.855134",
+  secondaryMenu: "1.865038",
+  image: "1.77257",
+  footer: "1.202278",
+  search: "1.77262",
+  language: {
+    en: "1.77273",
+    sv: "1.272446",
+  },
+  analytics: "1.464751",
+  gtmAnalytics: "1.714097",
+  gtmNoscript: "1.714099",
+};
+
+/** Renew a cached-block */
+async function renewCortinaBlock(str) {
+  log.info(`Renewing Cortina block ${str}`);
+
+  const body = got
+    .get(`https://www.kth.se/cm/${blocks[str]}`)
+    .then((res) => res.body);
+
+  await body;
+
+  // We store the Promise after getting the value, not the value
+  cachedBlocks[str] = body;
+}
+
+/** Returns the cached-version of the block or fetches it */
+async function fetchCortinaBlock(str) {
+  if (cachedBlocks[str]) {
+    log.debug(`Cache hit for Cortina block ${str}`);
+    return cachedBlocks[str];
+  }
+
+  // We don't await here. Instead we return the promise
+  const body = got
+    .get(`https://www.kth.se/cm/${blockIds[str]}`)
+    .then((res) => res.body);
+
+  cachedBlocks[str] = body;
+  return body;
+}
+
+module.exports = { compileTemplate, fetchCortinaBlock, renewCortinaBlock };
--- a/server/utils.js
+++ b/server/utils.js
@@ -36,29 +36,16 @@ async function renewCortinaBlock(str) {
   log.info(`Renewing Cortina block ${str}`);
 
   const body = got
-    .get(`https://www.kth.se/cm/${blocks[str]}`)
-    .then((res) => res.body);
-
-  await body;
-
-  // We store the Promise after getting the value, not the value
-  cachedBlocks[str] = body;
-}
-
-/** Returns the cached-version of the block or fetches it */
-async function fetchCortinaBlock(str) {
-  if (cachedBlocks[str]) {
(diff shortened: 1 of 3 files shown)


server/utils.js
```diff
@@ -10,4 +12,53 @@ function compileTemplate(resolvePath, name) {
...
+  log.info(`Renewing Cortina block ${str}`);
+
+  const body = got
+    .get(`https://www.kth.se/cm/${blocks[str]}`)
+    .then((res) => res.body);
+
+  await body;
+
+  // We store the Promise after getting the value, not the value
+  cachedBlocks[str] = body;
```
kaj, line 45: What do this actually do?  We await the body, but we don't do anything with the resolved value.  Instead we store the already resolved promise?  Why?  If we want a promise, why do we await it before storing it and returning?  If we want it resolved, why don't we just store the resolved body?

What happens if there is an error awaiting body?  Is that ignored with the discarded return value of await or does it raise an exception?

server/utils.js
```diff
@@ -10,4 +12,53 @@ function compileTemplate(resolvePath, name) {
...
+  footer: "1.202278",
+  search: "1.77262",
+  language: {
+    en: "1.77273",
+    sv: "1.272446",
+  },
+  analytics: "1.464751",
+  gtmAnalytics: "1.714097",
+  gtmNoscript: "1.714099",
+};
```
kaj, line 32: There's a lot more blocks defined here than those that are periodically renewed.  As the cache is just a dictionary and I don't see any periodic cleaning, any other of the blocks will be fetched once and then "cached" until the server terminates.  Is that wanted?

server/utils.js
```diff
@@ -10,4 +12,53 @@ function compileTemplate(resolvePath, name) {
...
+    return cachedBlocks[str];
+  }
+
+  // We don't await here. Instead we return the promise
+  const body = got
+    .get(`https://www.kth.se/cm/${blockIds[str]}`)
+    .then((res) => res.body);
+
+  cachedBlocks[str] = body;
+  return body;
```
kaj, line 61: This code is duplicated from `renewCortinaBlock`.  Couldn't we call that (and have it return the promise / body) on cache miss rather than duplicating it?

kaj (commented):
I've got some questions ...

server/utils.js
exacs: >What do this actually do? We await the body, but we don't do anything with the resolved value. Instead we store the already resolved promise? Why?

The goal of this function is to perform a renewal in the background. So if something else wants to get the block _while this is fetching a new version_, that old version of the block is available.

> If we want a promise, why do we await it before storing it and returning? If we want it resolved, why don't we just store the resolved body?

One problem I've seen with storing the body itself is _what happen if there is a request to /kpm when the block is fetched but we are awaiting for the response?_. Instead of triggering one more request, everything awaits for the same response.

> What happens if there is an error awaiting body? Is that ignored with the discarded return value of await or does it raise an exception?

It will raise an exception. I think they should be handled slightly different depending on:

- It happened when the cache is empty (should we retry?, should we leave it empty and retry when the next request comes?)
- It happen when renewing (i.e. there is something in the cache, we might want to just ignore the error)

server/utils.js
exacs: This "blockIds" is just a map of `keys` (strings) to "cortina ID"s. Only three of them are actually renewed or fetched the first time

kaj: Yes, that's my point.  Any of them could be used in a template, in which case we have a problem if we don't remember to also add it to the renew setup.

kaj:
After thinking some more, I want to suggest an alternative implementation, where `fetchCortinaBlock` only returns the already loaded string from the cache (it does not even need to be async), and _only_ the initializer / periodic update ever writes anything in the cache.

For the three blocks we handle, the only change is that `fetchCortinaBlock` would be marginally faster (for not being async and not checking the value).  For any as yet unhandled block it would make the block render empty until we add it to the cache setup, which makes it easier to not miss.

exacs:
I think I forgot to add a bit of explanation about the design of this cache. I had two things in mind:

1. When there is nothing in the cache (when app starts)
2. When there is something old in the cache (every hour after app start)

And for every point, what happens if a new request to KPM comes while the app is fetching a block.

---

For point 1, when the app starts, it fetches all blocks as soon as possible. [after starting the server](https://github.com/KTH/kpm/pull/16/files#diff-e07d531ac040ce3f40e0ce632ac2a059d7cd60f20e61f78268ac3be015b3b28fR15-R21) is what I have chosen.

- This function is `async` which means that the server is accepting requests even if a block is being downloaded.
- If a request comes when fetching is still going on, I decided to not perform extra requests. So, `fetchCortinaBlock` is designed in a way that _it never does two `fetch` to the same element at the same time_.

  That's the reason why `fetchCortinaBlock` returns a Promise and also the cache saves that promise:

  So, if a second call to `fetchCortinaBlock` is done while the first one is not finished (first fetch has started but not finished), the function returns the _same_ promise, meaning that, when the first fetch is finished, the second will finish aswell.

For point 2, I thought it was better that requests always has quick access to the content regardless if its old or not. Better have access to an old request but be fast.

- That's the reason why `renewCortinaBlock` awaits for the promise and only when its fulfilled, it is saved.

kaj (approved):

PR #17, opened 2021-01-27
implemented basic error handling

miyamoto-san:


server/panels/router.js
```diff
@@ -6,51 +6,51 @@ const path = require("path");
...
   }
 });
 
-function corsAllow(res, req) {
-  res.header("Access-Control-Allow-Origin", req.headers["origin"] || "*");
-  res.header("Access-Control-Allow-Credentials", "true");
-  res.header("Access-Control-Allow-Methods", "POST, GET, OPTIONS");
-  //res.header("Access-Control-Allow-Headers", "X-Requested-With");
-  res.header("Vary", "Origin");
-}
```
kaj, line 56: This function, and the calls to it, are still needed.  Or should those headers be set in some other way?

server/server.js
```diff
@@ -47,18 +47,33 @@ app.get("/kpm/logout", (req, res) => {
...
-      footer,
-      megaMenu,
-      search,
-    })
-  );
+app.get("/kpm", async (req, res) => {
+  try {
+    const footer = await fetchBlock("footer");
+    const megaMenu = await fetchBlock("megaMenu");
+    const search = await fetchBlock("search");
```
kaj, line 55: This also seems like a rebase error, as this reverts the caching PR by @exacs that was previously merged.

server/login-router.js
```diff
@@ -50,17 +50,22 @@ async function authenticateUser(req) {
...
+    const loginUrl = new URL(`${process.env.SSO_ROOT_URL}/login`);
+    loginUrl.searchParams.set("service", serviceUrl);
 
-  log.info(`Redirect user to ${loginUrl}`);
-  res.redirect(loginUrl);
+    log.info(`Redirect user to ${loginUrl}`);
+    res.redirect(loginUrl);
+  } catch (err) {
+    log.error(err);
+    return res.status(400).send("");
```
kaj, line 67: How is any unexpected exception caught in the handler functions a request error, while similarly unexpected exceptions caught in the catchAll are server errors?

kaj (changes requested):
Two things I think are rebase mistakes and one question about the actual error handling:

server/login-router.js
exacs: Agree here. I think unexpected exceptions should not be handled in routes but in the `catchAll`

server/panels/router.js
```diff
@@ -1,20 +1,25 @@
 const log = require("skog");
 const { Router } = require("express");
-const handlebars = require("handlebars");
-const fs = require("fs");
-const path = require("path");
 const { compileTemplate } = require("../utils");
 
 const panelsRouter = Router();
-module.exports = panelsRouter;
```
exacs, line 9: This was move down. I don't have any opinions on that :)

kaj (approved):
Much smaller fixes now.  :-)  Seems ok to me from reading the code.

PR #18, opened 2021-02-03
feat: added button in frontend

miyamoto-san:


server/log-router.js
```diff
@@ -0,0 +1,18 @@
+const log = require("skog");
+const { Router } = require("express");
+const { compileTemplate } = require("./utils");
+
+const logRouter = Router();
+
+const toggleMenuTemplate = compileTemplate(__dirname, "toggle-menu.handlebars");
+
+logRouter.get("/enabled", (_, res) => {
+  log.info("kpm_use cookie enabled");
```
miyamoto-san, line 10: Question: does this suffice? Or should we add some more info for logging here?

kaj: I would like to add the kthid (and there shoud be a logged in user doing this).  Preferably as a a data field. Can you do like this with skog?

```javascript
const kthid = getKthidFromReqest(req);
log.info("kpm_use cookie enabled", {"kthid": kthid});
```

kaj (commented):
As the `/kpm/log/...` request already does more than logging (it returns the htm for a new button), I think I would rename it to a `/kpm/set/...`, change it to a `POST` request (which would also avoid potential probe / preview request), and set the cookie with a response header rather than client-side code.

Thoughts?

server/log-router.js
exacs: Yes, it can be done with skog. It has exactly the same interface as bunyan (actually it calls bunyan under the hood). To be more precise, the syntax is:

```js
log.info({kthid: kthid}, "kpm_use cookie enabled")
```

From their documentation https://github.com/trentm/node-bunyan#log-method-api:

<details>
```js
log.info();     // Returns a boolean: is the "info" level enabled?
                // This is equivalent to `log.isInfoEnabled()` or
                // `log.isEnabledFor(INFO)` in log4j.

log.info('hi');                     // Log a simple string message (or number).
log.info('hi %s', bob, anotherVar); // Uses `util.format` for msg formatting.

log.info({foo: 'bar'}, 'hi');
                // The first field can optionally be a "fields" object, which
                // is merged into the log record.

log.info(err);  // Special case to log an `Error` instance to the record.
                // This adds an "err" field with exception details
                // (including the stack) and sets "msg" to the exception
                // message.
log.info(err, 'more on this: %s', more);
                // ... or you can specify the "msg".

log.info({foo: 'bar', err: err}, 'some msg about this error');
                // To pass in an Error *and* other fields, use the `err`
                // field name for the Error instance **and ensure your logger
                // has a `err` serializer.** One way to ensure the latter is:
                //      var log = bunyan.createLogger({
                //          ...,
                //          serializers: bunyan.stdSerializers
                //      });
                // See the "Serializers" section below for details.
```
</details>

miyamoto-san:
> As the `/kpm/log/...` request already does more than logging (it returns the htm for a new button), I think I would rename it to a `/kpm/set/...`, change it to a `POST` request (which would also avoid potential probe / preview request), and set the cookie with a response header rather than client-side code.
> 
> Thoughts?

Agreed in so far as router name goes. I've changed that and also made it a POST instead of a GET. As for whether the cookie is set on the backend or the frontend. What do you suggest is the advantage of doing it backend over frontend?

exacs:
> Agreed in so far as router name goes. I've changed that and also made it a POST instead of a GET. As for whether the cookie is set on the backend or the frontend. What do you suggest is the advantage of doing it backend over frontend?

I agree with @kaj. I think both _setting the cookie_ and _logging that the cookie is set_ should be close to each other (i.e. the same "function" that sets the cookie should log that the cookie is set). Specifically for the question about backend vs frontend:

- Frontend implementation would save round-trips to the server but in this case they are needed anyways in order to fetch the handlebars-powered HTML
- Backend implementation means less code in the client-side `kpm.js` bundle:
  - Better separation of concerns (the code is not used from the menu itself but from the "info page")
  - Smaller `kpm.js`

miyamoto-san:
> > Agreed in so far as router name goes. I've changed that and also made it a POST instead of a GET. As for whether the cookie is set on the backend or the frontend. What do you suggest is the advantage of doing it backend over frontend?
> 
> I agree with @kaj. I think both _setting the cookie_ and _logging that the cookie is set_ should be close to each other (i.e. the same "function" that sets the cookie should log that the cookie is set). Specifically for the question about backend vs frontend:
> 
>     * Frontend implementation would save round-trips to the server but in this case they are needed anyways in order to fetch the handlebars-powered HTML
> 
>     * Backend implementation means less code in the client-side `kpm.js` bundle:
>       
>       * Better separation of concerns (the code is not used from the menu itself but from the "info page")
>       * Smaller `kpm.js`

Sure, but what are we optimising towards? To me this looks like premature optimisation. The frontend file is no where near a problematic file size. If these are the only concerns I think we can go ahead an merge it. Round trips nor file size shouldn't be of any concern to us at this time, if ever.

miyamoto-san:
To clarify.

```js
response.writeHead(200, {
    'Set-Cookie': 'use_kpm=true',
    'Content-Type': 'text/plain'
  });
```
This would be an acceptable way of setting the cookie backend in my opinion. I don't want to introduce any new packages like `cookie-parser` or similar for something so arbitrary.

Now of course handling headers at this low level is quite bulky which is why I think we can safely opt in for doing it frontend, especially considering we've already got a reliable package for doing it there.

And as I said before I think round trips and file size optimisations are quite far ahead of us at this point. I suggest this will only truly be important at a stage in which this info page probably isn't even used any longer.

PR #19, opened 2021-02-10
Cookie options

miyamoto-san:
Added some cookie options

### DIFF
--- a/client/toggle-menu.js
+++ b/client/toggle-menu.js
@@ -8,6 +8,7 @@ const options = {
   ...(isProd && { secure: true }),
   expires,
   ...(isProd && { domain: ".kth.se" }),
+  sameSite: "Strict",
 };
 console.log({ options });
 async function logStatus(status) {
--- a/client/toggle-menu.js
+++ b/client/toggle-menu.js
@@ -1,10 +1,20 @@
 import Cookies from "js-cookie";
 
 const COOKIE_NAME = "use_kpm";
+const isProd = process.env.NODE_ENV === "production";
+const expires = 365 * 24 * 3600 * 1000;
 
+const options = {
+  ...(isProd && { secure: true }),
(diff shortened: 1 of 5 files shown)


kaj (commented):
One week might be a bit short for the cookie expire, I think permanent is good here?

Also, I don't se how this requires 10k new lines in package-lock.json (especially since there is no change at all in package.json).   A few sporadic updates is fine, but 10k new lines?  What's npm doing here?

exacs:
@kaj 

> Also, I don't se how this requires 10k new lines in package-lock.json (especially since there is no change at all in package.json). A few sporadic updates is fine, but 10k new lines? What's npm doing here?

Good catch! It seems that there is an explanation for this. You can see that the file has a new "lockfileVersion" number: (line 4)

```diff
-  "lockfileVersion": 1,
+  "lockfileVersion": 2
```

I am reading about this. This `v2` of the file is very different from `v1` according to [package-lock.json docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion):

> Note that the file format __changed significantly in npm v7__ to track information that would have otherwise required looking in node_modules or the npm registry. Lockfiles generated by npm v7 will contain lockfileVersion: 2.
>
> - No version provided: an "ancient" shrinkwrap file from a version of npm prior to npm v5.
> - 1: The lockfile version used by npm v5 and v6.
> - 2: The lockfile version used by npm v7, which is backwards compatible to v1 lockfiles.
> - 3: The lockfile version used by npm v7, without backwards compatibility affordances. This is used for the hidden lockfile at node_modules/.package-lock.json, and will likely be used in a future version of npm, once support for npm v6 is no longer relevant.
>
> npm will always attempt to get whatever data it can out of a lockfile, even if it is not a version that it was designed to support.

PR #21, opened 2021-03-15
Openid connect

exacs:
This PR replaces CAS with OpenID Connect

Still TODO before merging:

- [x] Handle redirect like the old way (with `next` parameter)
- [x] Implement logout
- [x] Remove CAS from code
- [x] Clear environmental variables to reduce duplication


server/server.js
```diff
@@ -40,13 +43,9 @@ app.get("/kpm/_monitor", (req, res) => {
...
-app.use("/kpm/login", loginRouter);
-app.get("/kpm/logout", (req, res) => {
-  req.session.destroy();
-  const logoutUrl = new URL(`${process.env.SSO_ROOT_URL}/logout`);
-  res.redirect(logoutUrl);
-});
-
+app.get("/kpm/auth/login", openid.redirectToLogin);
+app.post("/kpm/auth/callback", openid.processCallback);
+app.get("/kpm/auth/logout", openid.redirectToLogout);
```
kaj, line 48: Borde man inte kunna göra typ `app.use("/kpm/auth", openidRouter)` här?

server/server.js
```diff
@@ -1,17 +1,21 @@
...
 
 if (isDev) {
   log.info("App is in development mode");
 }
 
 const infoPageTemplate = compileTemplate(__dirname, "info-page.handlebars");
+const domain = new URL(process.env.SERVER_HOST_URL).hostname
+  .split(".")
+  .slice(-2)
+  .join(".");
```
exacs, line 18: Not very convinced about doing something like this to get the `domain`. If someone have better Ideas, please tell me! <3

kaj: I think this is fine.

kaj (approved):

PR #23, opened 2021-03-16
Load course names

kaj:
Load course name (in swedish and english) from kopps, with redis caching.

To do before merging:
- [x] Configure REDIS_URL though env (was: This just creates the redis client with default parameters.  Will that be correct in azure, or do we need to handle environment variables and pass them as parameters to redis `createClient()`?)
- [ ] This puts all of the kopps and redis code directly in `server/openid/extract-info.js`.  It's not too long (yet), but maybe it should be some other structure anyway?
- [x] Redis is a cache, so how do we specify how long time to cache things?


kaj:
Kanske man borde sätta en ännu högre `timeout` för kopps, men i stället sätta `retry` till `0`.

PR #25, opened 2021-03-19
Look at groups for registered courses.

kaj:


kaj:
Two questions regarding `lookupCanvasLinks`:

1. It looks up the course room by SISID for the course round.  That implies it will fail when the students are cross-listed into another course room (sammanslagna kursrum).  Would it be possible to look up through the sections instead?  Or maybe do like the old personal meny: fetch all the course room the current user has access to and try in more than one way to find the right course for each of them?
2. If there is an error, it is properly logged, but nothing is returned and the caller expect an array, so we get a new error in the caller.  So maybe just return an empty array after logging the error?

kaj:
I also think that first getting a list of mixed stuff and then splitting it up in `completedCourses` and `activeCourseRounds` and the separate handling of their `lookupCourseData` is too complicated.  But that may be the subject for a further PR after mergin this?

me97esn:
The first time I looked at studies for one of the test users with many courses, the menu displayed nothing except the text "undefined". But after a refresh it displays correct. A restart of the server don't re-produce the error. Perhaps it could be caused when the cache is empty, but when the data exist in the cache it displays correctly?

me97esn:
I noticed another issue: when the menu too big to be displayed on the screen, it's not possible to view the content at the bottom. If I scroll down, it scrolls the page in the background instead of the menu;

kaj:
> The first time I looked at studies for one of the test users with many courses, the menu displayed nothing except the text "undefined".

Yes, this is a known problem that just got a little easier to expose.  :-)  If a user tries to open the menu panel before the group parsing is done, the panel should wait for it.  It needs to be fixed, but maybe not as part of this PR.

> I noticed another issue: when the menu too big to be displayed on the screen, it's not possible to view the content at the bottom.

Yes, the css needs some work.  But I think that is a separate PR as well.  :-)

me97esn:
image
I tested again this morning, after the Redis cache has expired, and had the same problem again. No studies displayed, only `undefined`.

server/openid/extract-info.js
```diff
@@ -44,21 +51,159 @@ async function lookupCourseData(courseCode) {
...
+
+    await cache.set(key, JSON.stringify(data), "EX", cache.expire);
+    return [data];
+  } catch (error) {
+    log.error({ sisId, err: error }, "Failed to get canvas information");
+    return [];
+  }
+}
+
+function parseUgGroup(name) {
```
me97esn, line 80: IMO this function should be refactored. The name suggests that the function just extracts some data from a name, but when looking at the code it does a few different things:
- Checks if it is course or a course round
- Checks if it is godkänd
- Checks if it's registrerad or antagen (and handles them differently)

Imo a function should have _one_ clear responsibility, but it's difficult to decide upon what's the responsibility of this function. I also think that any function should aim for being small enough so you don't have to scroll to see the entire function (which I know only work in some cases, but it's still something to aim for if possible).

If it's possible I would prefer if we could split it into a few smaller functions, with one responsibility each

server/openid/extract-info.js
```diff
@@ -44,21 +51,159 @@ async function lookupCourseData(courseCode) {
...
+    .filter((group) => group);
 
-  for (const group of token.memberOf) {
-    const match = COMPLETED_COURSE_REGEX.exec(group);
+  const completedCourses = parsedGroups.filter(
+    (group) => group.courseCode && !group.sisId
+  );
+  const activeCourseRounds = parsedGroups.filter(
+    (group) => group.courseCode && group.sisId
+  );
```
me97esn, line 183: This logic isn't obvious for me, and takes some time to read before I understand. I would prefer if the data `parsedGroups` had a boolean: `isCompleted` or a function `isCompleted()`, which is a lot more obvious then if a course has a sisId or not

server/openid/extract-info.js
```diff
@@ -44,21 +51,159 @@ async function lookupCourseData(courseCode) {
   }
 }
 
+async function lookupCanvasLinks(courseRound) {
```
me97esn, line 54: Perhaps this function and lookupCourseData should be refactored, since both use similar logic:
1) get a key from the cache, and returns if it exists
2) If the data doesn't exist in the cache, fetch it somewhere else
3) Store the data in the cache.

We should be able to use the same code for 1 and 3 above, but different code for 2

server/openid/extract-info.js
```diff
@@ -44,21 +51,159 @@ async function lookupCourseData(courseCode) {
...
+
+    const { body } = await canvas.get(`courses/sis_course_id:${sisId}`);
+
+    const data = {
+      url: `https://canvas.kth.se/courses/${body.id}`,
+      published: body.workflow_state === "available",
+    };
+
+    await cache.set(key, JSON.stringify(data), "EX", cache.expire);
+    return [data];
```
me97esn, line 73: Why is this wrapped in an array? The call on line 65 always returns one or zero canvas courses, not multiple. What's the logic in wrapping it in an array?

server/panels/router.js
```diff
@@ -56,9 +56,14 @@ panelsRouter.get("/studies", (req, res) => {
   log.info("Requesting panel '/studies'");
```
me97esn, line 56: Do I understand this function correctly when I say that it's a get (triggered by a HTTP GET method) that changes data stored in the session? Since the session object isn't immutable. Is this on purpose, or just a mistake? HTTP GET shouldn't change anything, GET shouldn't have side effects.
I suggest rebuilding this function so that it creates a copy of the data in the session, changes the data within that copy only, and passes that data to the `studiesPanel()` function. So that the session data is immutable unless we explicitily change the session data.
If this function _should_ change the session data, I think it should both be obvious that it changes session data, be called with either a POST or a PUT, and be named in some way to describe what it actually does (updateSessionLinks or something similar)

me97esn: The problem with using a language where nothing is immutable, I know :)

server/panels/router.js
```diff
@@ -68,6 +73,7 @@ panelsRouter.get("/studies", (req, res) => {
           }
         }
       }
+      course.status = [...status.keys()].join(" ").toLowerCase();
```
me97esn, line 76: This line takes an array of statuses and converts it into a string. IMHO this is better handled in the client side (or in the server side handlebars template) instead of in the api. I would prefer if this would just be `course.statuses = status` and the rest is done in the handlebars template. This way we design this file as an api that _could_ be called by the browser if we want to change the way we build the frontend some time in the future and start using React for instance.

server/panels/router.js
exacs: No, I don't think so. I think this function does NOT change the data in session. It reads the data from the session but doesn't change it.

Session is updated when the user logs in not when this panel is opened.

server/panels/router.js
exacs: This is left as it is to be compatible with the current handlebars template. Should the PR include a better hbs template or should be in a different PR?

server/openid/extract-info.js
exacs: Hmm. Sure it can be divided into different functions but remember that all the functions are called only ONCE. 

It is better to scroll through one big function or to scroll up-and-down through 4 functions splitted in different parts?

Sure this function can be refactored or redesigned :)

server/panels/router.js
me97esn: I tested it, and can verify that it indeed changes the session data. A bit difficult to test since it requires a user with active courses. But: if a user has an active course, the `req.session.userData.activeStudentCourses` has no `canvasLinks` property _before_ this function is called, but it does _after_ this function is called.

exacs: Are you sure that is because this function is called? As I remember, the session object is filled when the user logs in but not immediately (it can take some seconds). Maybe you have open the tab before the session object was fully updated...

exacs: You are right. I've checked now code and yes, the session object is updated within this function :)

me97esn (changes requested):
At least the mutation of the session should be fixed before this PR is merged. The rest of the comments are open for discussion.

kaj:
This is for the old code, and not directly relevant to what we write in 2022.

PR #38, opened 2022-10-07
Feature support page transitions for mobile

jhsware:
We now have a menu with routing and animated menus.


jhsware:
Chose a different route

PR #43, opened 2022-10-18
Make my-teaching-api return roles per course.

kaj:
Not ready to merge yet, see review comments.

### DIFF
--- a/apis/my-teaching-api/src/api.ts
+++ b/apis/my-teaching-api/src/api.ts
@@ -37,11 +37,6 @@ export type TUgUser = {
 };
 
 export type TUgGroup = {
-  description: {
-    sv: string;
-    en: string;
-  };
-  kthid: string;
   name: string;
 };
 
@@ -75,9 +70,26 @@ api.get("/user/:user", async (req, res, next) => {
   if (json === undefined || statusCode !== 200) {
     if (IS_DEV) {
       return res.status(statusCode || 500).send(data);
+    } else {
+      return res.status(statusCode || 500).send("error");
     }
+  } else {
+    const result = teachingResult(json);
+    res.status(statusCode || 200).send(result);
   }
+});
 
+export type Role = {
+  role: string;
+  year: string;
+  term: string;
+  round_id: string;
+};
+
+// Separate function for testability
+export function teachingResult(data: TUgGroup[]): {
+  [index: string]: Array<Role>;
+} {
   /* nameRegex regex test matches:
       Match: edu.courses.5B.5B1219.examiner
       No match: edu.courses.5B.5B1219.examiner.sdf
@@ -91,8 +103,8 @@ api.get("/user/:user", async (req, res, next) => {
   */
   const nameRegex =
     /^edu\.courses\.[^.]+\.(?<course_code>[^.]+)\.((?<role>examiner)|(?<year>[0-9]{4})(?<term>[0-9])\.(?<round_id>[^.]+)\.(?<role_alt>teachers|courseresponsible|assistants))$/i;
-  const result = json
-    ?.map((o) => o.name.match(nameRegex)?.groups)
+  const result = data
+    .map((o) => o.name.match(nameRegex)?.groups)
     .filter((o) => o)
     .map((o: any) => {
       const { role, role_alt, ...other } = o;
@@ -101,6 +113,14 @@ api.get("/user/:user", async (req, res, next) => {
         role: role || role_alt,
       };
     });
-
-  res.status(statusCode || 200).send(result);
-});
+  let courses: { [index: string]: Array<Role> } = {};
+  for (const { course_code, ...role } of result) {
+    console.log(`Got course ${course_code} with roles ${role}`);
+    if (courses[course_code]) {
+      courses[course_code].push(role);
+    } else {
+      courses[course_code] = [role];
+    }
+  }
+  return courses;
+}
--- a/apis/my-teaching-api/__test__/callDummyUser.spec.ts
+++ b/apis/my-teaching-api/__test__/callDummyUser.spec.ts
@@ -1,6 +1,6 @@
 import { describe, expect, jest, test } from "@jest/globals";
 import { UGRestClient } from "kpm-ug-rest-client";
-import { TUgUser } from "../src/api";
+import { TUgUser, teachingResult } from "../src/api";
 import { addMockResponse } from "../__mocks__/openid-client";
 
 jest.mock("openid-client");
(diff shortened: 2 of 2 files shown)


apis/my-teaching-api/src/api.ts
```diff
@@ -5,12 +5,12 @@ import { UGRestClient } from "kpm-ug-rest-client";
...
-const CLIENT_ID = process.env.CLIENT_ID;
-const CLIENT_SECRET = process.env.CLIENT_SECRET;
+const CLIENT_ID = process.env.CLIENT_ID || "wtf";
+const CLIENT_SECRET = process.env.CLIENT_SECRET || "wtf";
 const UG_REST_BASE_URI = process.env.UG_REST_BASE_URI || "https://integral-api.sys.kth.se/test/ug";
 
-assert(typeof CLIENT_ID === "string" && CLIENT_ID, "Missing CLIENT_ID for OpenID auth");
-assert(typeof CLIENT_SECRET === "string" && CLIENT_SECRET, "Missing CLIENT_SECRET for OpenID auth");
+// assert(typeof CLIENT_ID === "string" && CLIENT_ID, "Missing CLIENT_ID for OpenID auth");
+// assert(typeof CLIENT_SECRET === "string" && CLIENT_SECRET, "Missing CLIENT_SECRET for OpenID auth");
```
kaj, line 13: I think these asserts should be moved to `index.ts`, but then the above variable declarations also needs to be moved.  Without these asserts, the type of `proces.env.FOO` is `string | undefined`.  That's why i temporarily added `|| "wtf"` above to make sure they are strings and keep the type system happy.

apis/my-teaching-api/src/api.ts
```diff
@@ -38,11 +38,6 @@ export type TUgUser = {
 }
 
 export type TUgGroup = {
-  description: {
-    sv: string,
-    en: string
-  },
-  kthid: string,
   name: string,
 }
```
kaj, line 42: I don't think the description or kthid of an UG group are relevant to us, so even if they exists in the data, I see no reason for them to exist here (and keeping them makes mocking more verbose).

apis/my-teaching-api/src/api.ts
```diff
@@ -76,9 +70,24 @@ api.get("/user/:user", async (req, res, next) => {
   if (json === undefined || statusCode !== 200) {
     if (IS_DEV) {
       return res.status(statusCode || 500).send(data);
+    } else {
+      return res.status(statusCode || 500).send("error");
```
kaj, line 74: This is just to always send something, since I moved the rest of this function into an else to make sure that `json` is actually defined there.

apis/my-teaching-api/src/api.ts
```diff
@@ -76,9 +70,24 @@ api.get("/user/:user", async (req, res, next) => {
...
+
+export type Role = {
+  role: string,
+  year: string,
+  term: string,
+  round_id: string,
+}
 
+// Separate function for testability
+export function teachingResult(data: TUgGroup[]): { [index: string]: Array<Role> } {
```
kaj, line 90: I make the mapping from ug rest answer to our answer a separate function just so it can be tested.

kaj (commented):
I have some questions ...

apis/my-teaching-api/src/api.ts
jhsware: I suggest adding [@kth/reqvars](https://www.npmjs.com/package/@kth/reqvars) and removing the asserts entirely.

apis/my-teaching-api/src/api.ts
jhsware: Perhaps we could change this to:
`
return res.status(statusCode || 500).send({ "error": "No valid JSON returned from UG REST API." });
`

apis/my-teaching-api/src/api.ts
```diff
@@ -91,13 +100,21 @@ api.get("/user/:user", async (req, res, next) => {
...
+  let courses: { [index: string]: Array<Role> } = {};
+  for (const { course_code, ...role } of result) {
+    console.log(`Got course ${course_code} with roles ${role}`);
+    if (courses[course_code]) {
+      courses[course_code].push(role);
+    } else {
+      courses[course_code] = [role];
+    }
+  }
+  return courses
```
jhsware, line 119: Minor: we use semicolon terminated lines. Perhaps run prettier on the PR.

jhsware (changes requested):
I think we should fix the asserts, either move them or implement @kth/reqvars and remove them. Also implement the same solution for my-studies-api, my-canvas-rooms for consistency.

jhsware:
I will fix the broken kpm-test.

apis/my-teaching-api/src/api.ts
kaj: Reqvars seems to only check, not parse (or even get) the values?  The asserts here is mainly changing the type of those variables from `string | undefined` (or even `any`?) to `string`.

apis/my-teaching-api/src/api.ts
kaj: Yes.  Perhaps even changing the inner `if` to just log and send a known error?  Or even just to raise an exception if we check for that somewhere?

apis/my-teaching-api/src/api.ts
kaj: Prettier done (but I guess we should add a check to ci as well).

apis/my-teaching-api/src/api.ts
jhsware: Since all env-vars are strings the reqvars check will be equivalent to the assert so nothing lost. Also, the main use case is to answer the question "Why won't my app start?" when you check out the project for the first time.

apis/my-teaching-api/src/api.ts
jhsware: Going forward we probably want to throw errors so we use the main error handler (hooked up in index.ts) so we can enforce consistent error reporting. Perhaps we should mob-program error handling later this week and let this be a temporary fix.

apis/my-teaching-api/src/api.ts
jhsware: We're going to hate that check though... unless this was a traumatic experience I think we should leave it as is for now.

apis/my-teaching-api/src/api.ts
kaj: For the "why won't it start", reqvars is as good as this.  But for the types, I don't think reqvars (in index.ts) can tell the usage here that those specific environment variables are defined.  Note that just commenting out the checks made type checking fail in api.ts, so I had to add those temorary `|| "wtf"` parts.

jhsware:
@kaj the CD/CI-pipeline is fixed in main so you can rebase to get your tests to pass.

jhsware (approved):

PR #44, opened 2022-10-19
Try to run prettier as an action.

kaj:
For now, this is just a suggestion on how to run pettier in github actions, as a separate job.  It could also be added to the test step in each package, so evolene would run it, but I think it is cleaner like this.  Thoughts?

Also, I renamed the "Continous Integration" job to "CI", to make room for showing the workspace name (matrix parameter) in the Actions view.  Ok?

Before merging:
- [x] Rebase on the fix for the build error in kpm.
- [ ] Actually run prettier and commit the fixes.


jhsware:
@kaj the CD/CI-pipeline is fixed in main so you can rebase to get your tests to pass.

kpm/kpm-frontend/src/panes/studies.tsx
```diff
@@ -19,21 +17,25 @@ export function Studies() {
...
 }
 
 function Course({ courseCode, year, status }: any) {
   return (
     <li className="kpm-studies-course">
       <h2>{courseCode}</h2>
-      <h3>{year} | {status}</h3>
+      <h3>
+        {year} | {status}
+      </h3>
```
kaj, line 38: This fells unneccessarily multi-linish.

kaj (commented):
Maybe we should do some configuration of prettier?

PR #50, opened 2022-10-27
Entry js

exacs:
This PR implements a new route `/kpm/widget.js` that will send different JS files depending on users' `use_kpm` cookie and log in status. In case user wants to use kpm and is logged in, it returns a JS that injects the artifacts generated by Parcel.

---

Note: this PR does not implement the response when user has not `use_kpm` nor is logged out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 node_modules
 .parcel-cache
 kpm/kpm-frontend/dist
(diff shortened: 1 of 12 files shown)


jhsware:
Call `npm run dev:prod` and go to http://localhost:3000/kpm/index.html to test.

PR #56, opened 2022-11-14
Improve dropdown and modal behaviour

jhsware:


jhsware:
Merging after concept review with Rasmus

PR #66, opened 2022-12-01
Protect APIs

exacs:
This PR adds authorization logic between KPM backend and my-xxx-apis.

- Declare env variables for access tokens
- Protect my-xxx-api behind a token
- Send authorization tokens from kpm to my-xxx-api


jhsware (changes requested):
Could we split the api routes into open and protected routes and move the token enforcement into index.ts as a middleware so we get something like this:

```
app.use(PREFIX, apiOpen);
app.use(PREFIX, requireTokenMiddleware, apiProtected);
```

Then we eliminate the risk of forgetting to protect API-routes if we add endpoints in the future.

PR #71, opened 2022-12-05
Fix kpm so it doesn't cover the full page

exacs:
This PR fixes the `inset` (top, bottom, left, right) CSS properties so that KPM does not cover the full page

(Note: this PR might include more small CSS fixes)

### DIFF
--- a/kpm/kpm-frontend/src/app.scss
+++ b/kpm/kpm-frontend/src/app.scss
@@ -6,7 +6,7 @@
   top: 0;
   left: 0;
   right: 0;
-  bottom: 0;
+  bottom: auto;
   z-index: 999;
 
   display: flex;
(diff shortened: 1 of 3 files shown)


jhsware (changes requested):
This will prevent the backdrop from covering the entire viewport. Not sure if there are other side effects too, but these needs to be looked at.

Another options is setting: pointer-events: none; on:

- #kpm-6cf53 

and then pointer-events: all; on:

- #kpm-6cf53 .kpm-menu
- #kpm-6cf53 .kpm-modal-backdrop
- #kpm-6cf53 .kpm-modal

I think these are the ones that are affected by the bottom: 0; change too.

jhsware (approved):

PR #110, opened 2023-01-24
Don't push entries to nav stack when navigating in menu

jhsware:
Avoid creating a long navigation history when interacting with the menu.

- Closing a pane should pop the entry in nav stack
- Switching pane should replace the entry in nav stack
- Clicking background should pop entry in nav stack


kaj (approved):
Looks good to me.  I assume the `navigate()` is part of react, and not from the browser, so there should be no compat issues, right?

jhsware:
Correct! Although technically it is from `react-router`.

PR #111, opened 2023-01-25
Change auth logging

jhsware:
To test:
1. Swap `new AuthError` for `new MutedAuthError` [on this line](https://github.com/KTH/kpm/blob/5652e29771251cf86bc81f857dda2cab00a309a0/kpm/kpm-backend/src/auth.ts#L155)
2. Run `npm run dev-activation`
3. Make a POST to http://localhost:3000/kpm/auth/callback?nextUrl=/doesNotMatter
 
You will now se that the error is logged as a warning and not error. This method is applied for `errors.OPError` further up in the same function and will suppress alerts for those non-critical errors during negotiation with the auth server.


kaj (approved):
LGTM.  Nice that you also fixed the host replacement in dev-activation!

PR #113, opened 2023-01-27
Improve error handling in my-canvas-rooms-api.

kaj:
Is this a good way handle errors that should not be status 500 (and does not need to be logged as error) ?

### DIFF
--- a/apis/my-canvas-rooms-api/src/api.ts
+++ b/apis/my-canvas-rooms-api/src/api.ts
@@ -1,6 +1,7 @@
 import express, { NextFunction, Request, Response } from "express";
 import CanvasClient, { CanvasRoom, CanvasApiError } from "./canvasApi";
 import log from "skog";
+import { MutedOperationalError } from "kpm-api-common/src/errors";
 
 export const api = express.Router();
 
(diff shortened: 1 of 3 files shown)


apis/my-canvas-rooms-api/src/api.ts
```diff
@@ -19,12 +20,28 @@ api.get(
...
+  constructor() {
+    super({
+      name: "EndpointError",
+      statusCode: 404,
+      type: "Not found",
+      message: "Not found in Canvas.",
+      details: undefined,
+    });
+  }
+}
```
jhsware, line 43: There is a specific error class called `EndpointError` that would be more appropriate here than MutedOperationalError. The idea is that an EndpointError is used when:
- the consumer should handle this error specifically in the area of code calling this endpoint
- the type string should be documented through the passed ErrType generic (ref APIAuthErrType in kpm-backend-interface/src/index.ts) in an interface.ts file or package
- the type should be of the PascalCase form

apis/my-canvas-rooms-api/src/canvasApi.ts
```diff
@@ -28,9 +29,20 @@ function getToken(token = "") {
     return token.substring(7);
   }
 
-  throw new Error(
-    "Unauthorized. You must access this endpoint either with a session or an authorization header"
-  );
+  throw new AuthTokenMissing();
+}
+
+class AuthTokenMissing extends MutedOperationalError<String> {
```
jhsware, line 35: There is a specific `AuthError` that would be more appropriate than MutedOperationalError. The ErrType generic should also be documented in interface.ts as above.

jhsware (changes requested):

kaj (commented):
👍🏻 Looks ok to me.  And I believe the issues you had earlier was fixed by your commit?  I have rebased, so this should be mergeable now?

(I can't submit an approving review since the original changes in the PR was from me, but I approve your changes.)

jhsware (approved):
LGTM

Note: jag tror jag ska kika igenom våra Exceptions igen. Jag gillar idén att man inte kan välja "fel" err type genom att subklassa EndpointError per projekt, men frågan är om man det är overkill att ha specifika Exceptions för varje case. Jag pingar dig för diskussion när jag kikar på det.

PR #130, opened 2023-02-16
When posting json, use json content-type.

kaj:
Plus convenience function `postApi` in kpm-frontend.


kpm/kpm-frontend/src/panes/utils.ts
```diff
@@ -74,6 +74,23 @@ export function formatTerm(startTerm: string) {
...
 
+export async function postApi(
+  path: string,
+  json: any,
+  options: RequestInit = {}
+): Promise<Response> {
+  const { headers, ...otherOptions } = options;
+  return await fetchApi(path, {
+    headers: {
+      "Content-Type": "application/json",
```
jhsware, line 85: Passing "Content-Type" in headers will overwrite our "Content-type" and given that we perform the JSON.stringify perhaps we should switch row 85 and 86 so we always force the content type to be correct?

jhsware: Yes, this is nitpicking :D

jhsware (changes requested):
Minor suggestion.

kpm/kpm-frontend/src/panes/utils.ts
kaj: Beeing able to overwrite the content-type is a feature, if we should talk to api:s that uses things like `application/vnd.myapplication.v1+json` for semantis and versioning.

jhsware (approved):
LGTM

PR #131, opened 2023-02-17
Mute CORS errors and make sure they send the correct statusCode

jhsware:
[403 Forbidden](https://en.wikipedia.org/wiki/HTTP_403)
The request contained valid data and was understood by the server, but the server is refusing action. This may be due to the user not having the necessary permissions for a resource or needing an account of some sort, or attempting a prohibited action (e.g. creating a duplicate record where only one is allowed). This code is also typically used if the request provided authentication by answering the WWW-Authenticate header field challenge, but the server did not accept that authentication. The request should not be repeated.

I added "CorsError" as possible error name because it is such a specific kind of error of high magnitude.

### DIFF
--- a/kpm/kpm-backend/src/index.ts
+++ b/kpm/kpm-backend/src/index.ts
@@ -13,6 +13,8 @@ import log from "skog";
 import { widgetJsHandler, widgetJsAssets, previewHandler } from "./widget.js";
 import cors from "cors";
 import { activation } from "./activation";
+import { MutedOperationalError } from "kpm-api-common/src/errors";
+import { TCorsError } from "kpm-backend-interface";
 
 const IS_DEV = process.env.NODE_ENV !== "production";
 const IS_STAGE = process.env.DEPLOYMENT === "stage";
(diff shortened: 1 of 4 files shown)


kpm/kpm-backend/src/index.ts
```diff
@@ -110,7 +135,13 @@ if (!IS_DEV) {
         if (!origin || corsWhitelist.indexOf(origin) !== -1) {
           callback(null, true);
         } else {
-          callback(new Error(`Not allowed by CORS (origin: ${origin})`));
+          callback(
+            new CorsError({
+              message: `Not allowed by CORS (origin: ${origin})`,
+              type: "InvalidOrigin",
```
kaj, line 141: Since `type` must be a `TCorsError` and that only allows the value `"InvalidOrigin"`, maybe change the constructor to just set that and not take a type as an argument?  Maybe even let the constructor create the message from origin in the details, so the call here can be just `new CorsError({origin})` ?

jhsware: I was considering that we might want to have different results. If a CorsError is generated for a subdomain of kth.se we might want to flag it as a different type so we easily can spot sites that should be added to the white list

PR #133, opened 2023-02-21
Package and decorate error in memoizer for better general error logging

jhsware:
We could improve error handling in the actual resolver that causes the error, but this is the general error handler and will do for now.

### DIFF
--- a/kpm/kpm-backend/src/api/commonCache.ts
+++ b/kpm/kpm-backend/src/api/commonCache.ts
@@ -63,7 +63,7 @@ export function memoized<T extends TCacheable>({
 
     // Cache miss, calculate value, store and return
     const newValue: T | undefined = await fn(key).catch((err: any) => {
-      err.details = { key };
+      err.details = { ...err.details, memoized_key: key };
       log.error(err);
       return undefined;
     });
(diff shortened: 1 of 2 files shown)


kpm/kpm-backend/src/api/commonCache.ts
```diff
@@ -62,13 +62,12 @@ export function memoized<T extends TCacheable>({
...
     // Cache miss, calculate value, store and return
-    let newValue: T | undefined;
-    try {
-      newValue = await fn(key);
-    } catch (err: any) {
-      Error.captureStackTrace(err);
-      log.error(err, "Missing error handling in memoized function");
-    }
+    const newValue: T | undefined = await fn(key).catch((err: any) => {
+      err.details = { key };
```
kaj, line 66: Hm.  What if `err` has other details?  Mabye either just set `err.memoized_key = key` to make conflicts less probable (and avoid a single-field dictionary), or do something like:

```ts
   err.details |= {};
   err.details["memoized_key"] = key;
```

?

kaj (commented):
Looks good.  I have one question:

kpm/kpm-backend/src/api/commonCache.ts
jhsware: Fixed in e08918a05

kaj (approved):
LGTM.  Som vanligt tar du mitt ändringsförslag och gör som jag säger fast lite snyggare.  😎

PR #140, opened 2023-03-01
Show fav canvas rooms

jhsware:
If we have favs in Canvas we should show them, and only them, in the short list.

Fixed hard coded term selection

### DIFF
--- a/kpm/kpm-frontend/src/panes/teaching.tsx
+++ b/kpm/kpm-frontend/src/panes/teaching.tsx
@@ -72,9 +72,9 @@ function Course({ courseCode, course }: TCourseProps) {
   const aboutCourseUrl = `https://www.kth.se/kurs-pm/${courseCode}/om-kurs-pm`;
   // TODO: These should be changed to course rooms, check backend
   const {
-    current = [],
+    shortList = [],
     exams = [],
-    other = [],
+    allCourseRooms = [],
   } = course.rooms ? filterCanvasRooms(course.rooms) : {};
   const currentTerm = getCurrentTerm();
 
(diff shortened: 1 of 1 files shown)


kpm/kpm-frontend/src/panes/teaching.tsx
```diff
@@ -72,11 +72,11 @@ function Course({ courseCode, course }: TCourseProps) {
...
   // TODO: These should be changed to course rooms, check backend
   const {
-    current = [],
+    shortList = [],
     exams = [],
-    other = [],
+    allCourseRooms = [],
   } = course.rooms ? filterCanvasRooms(course.rooms) : {};
-  const currentTerm = "HT2022";
+  const currentTerm = getCurrentTerm();
```
kaj, line 79: Unrelated (only used by CourseAdminDropDown?, so it confused me a little) but good.  😎

kpm/kpm-frontend/src/panes/teaching.tsx
```diff
@@ -98,9 +98,9 @@ function Course({ courseCode, course }: TCourseProps) {
           {course.rooms?.length === 0 && (
             <p className="kpm-muted-text">{i18n("No rooms found in Canvas")}</p>
           )}
-          <CanvasRoomShortList rooms={current} />
+          <CanvasRoomShortList rooms={shortList} />
           <CanvasRoomExpandedList
-            rooms={[...current, ...other]}
+            rooms={allCourseRooms}
```
kaj, line 103: I guess this change is because shortList may contain exam rooms (if favorited), and those should not appear in the exapanded list?

kaj (commented):
Looks good (from looking at the code, I haven't tested it).  I have some questions to make sure I understood correctly.

At first I thought the `hasFavorite` was unneccessary, we could just do `shortList: rooms.filter((c) => c.favorite) || otherShortlist`, but since the other handling depends on `courseRooms.length`  that would be more complicated.  So this way is fine.

kpm/kpm-frontend/src/panes/teaching.tsx
jhsware: Correctamundo!

kpm/kpm-frontend/src/panes/teaching.tsx
jhsware: Affirmative!

PR #141, opened 2023-03-01
Use accept-language.

kaj:
Set accept-language on all backend requests in frontent, and propagate that header to any requests to social in the backend.

This fixes the translations for service links.

### DIFF
--- a/kpm/kpm-backend/src/session.ts
+++ b/kpm/kpm-backend/src/session.ts
@@ -65,9 +65,12 @@ export async function sessionApiHandler(
   try {
     const user = sessionUser(req.session);
     if (user) {
-      let notes = await getSocial<NumNotes>(user, "notifications/count").catch(
-        socialErr
-      );
+      const lang = req.headers["accept-language"];
+      let notes = await getSocial<NumNotes>(
+        user,
+        "notifications/count",
+        lang
+      ).catch(socialErr);
       logger.info({ notes, kthid: user.kthid }, "Notes according to social");
       user.numNewNotifications = notes?.new;
     }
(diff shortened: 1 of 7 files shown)


kpm/kpm-backend/src/session.ts
```diff
@@ -65,9 +65,12 @@ export async function sessionApiHandler(
...
     if (user) {
-      let notes = await getSocial<NumNotes>(user, "notifications/count").catch(
-        socialErr
-      );
+      const lang = req.headers["accept-language"];
+      let notes = await getSocial<NumNotes>(
+        user,
+        "notifications/count",
+        lang
+      ).catch(socialErr);
```
jhsware, line 73: Is this endpoint language aware? Or are you passing lang for consistency?

jhsware (changes requested):
Do we need to look at KOPPS-calls too?

kpm/kpm-backend/src/session.ts
kaj: Just for consistency.  Currently, that endpoint only returns an object-wrapped json number.

kaj:
The kopps-dependant things we cache are independent of language (contains both languages), and I think we should keep it that way.  So no, I don't think we should change our kopps calls. But possibly our backend calls (which use those cached values) ?

kpm/kpm-backend/src/session.ts
kaj: (I made the argument required, to avoid missing it anywhere it might be needed).

kaj:
Den kopps-endpoint vi använder i nuläget (`${KOPPS_API}/course/${course_code}/courseroundterms`) ger språkobjekt {sv: ..., en: ...}, så det är såna vi cachar.  Så inga problem där.  Vi skulle ju kunna tolka de objekten i kpm-backend och bara skicka rätt språk till kpm-frontend, men jag tycker nog att det kan vara en separat fråga till någon annan gång ...

jhsware:
LGTM

jhsware (approved):
LGTM

PR #142, opened 2023-03-02
Single values of memberOf

jhsware:
Single values of memberOf are returned as a string from login server but we expect an array.


kaj (approved):
LGTM

PR #143, opened 2023-03-02
Cleanup of SASS imports

jhsware:


kaj (commented):
Still a lot of `"./something"` that could be just `"something"` if you want to clean up some more.  And I believe even the `"../something"` can be changed to just "../something", as the something resides in the root scss path, which will be searched after the cwd of the file the `@use` statement is in.

But as is, it it still a good improvement, so feel free to merge with or without the extra cleanup!  😎

PR #144, opened 2023-03-06
Include all relevant registration codes.

kaj:
For course rooms that are combinded for multiple coures rounds, include all the relecant registration codes rather than one random.


jhsware (approved):
LGTM

PR #145, opened 2023-03-07
Rapp courses that end upp in shortlist should not be exluded

jhsware:


kaj (approved):
LGTM.  The filtering is done elsewhere, this should just render the list.

PR #147, opened 2023-03-07
Use data urls for svg icons.

kaj:
Use the `$data-something-svg` variables provided by kth-style rather than trying to load icons externally from the kth-style app.

This gets rid of the CSP error where we try to load ref urls in prod.

kmp-frontend already had a dependency to kth-style, so I only needed to use it.

### DIFF
--- a/kpm/kpm-frontend/src/_vars.scss
+++ b/kpm/kpm-frontend/src/_vars.scss
@@ -1,13 +1,8 @@
+@forward "kth-style/public/sass/variables/svg_data";
+
 // TODO: We need to go through these breakpoints,
 // also check headers on site for mobile pixel size
 $mblBreakpoint: "screen and (max-width: 430px)";
 $desktopBreakpoint: "screen and (min-width: 431px)";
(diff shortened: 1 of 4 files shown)


kpm/kpm-frontend/src/_vars.scss
```diff
@@ -1,13 +1,8 @@
+@forward "kth-style/public/sass/variables/svg_data";
```
kaj, line 1: I'd like the path here to be just `"kth-style/variables/svg_data"`, but it seems parcel just adds the entire `node_modules` directory to the load path without any kind of mapping.  Maybe either our use of parcel to kth-styles packaging could be improved?

jhsware: I guess you could pass [load paths](https://sass-lang.com/documentation/at-rules/use#load-paths) but I think it will cause more issues than it solves. I believe `dart-sass` is responsible for resolving these imports. Parceljs will only resolve references from TypeScript files.

jhsware (approved):
LGTM

PR #148, opened 2023-03-07
Use kth-style colors

kaj:
Use kth-style sass variables in stylesVars.scss, to get proper kth-style colors and document where our colours come from.

### DIFF
--- a/kpm/kpm-frontend/src/app.scss
+++ b/kpm/kpm-frontend/src/app.scss
@@ -133,7 +133,7 @@
 
   .kpm-muted-text {
     padding-top: 0.1em;
-    color: var(--kpmPaneTextSoft);
+    color: var(--kpmPaneTextMuted);
     font-size: 0.9em;
   }
 }
--- a/kpm/kpm-frontend/src/app.scss
+++ b/kpm/kpm-frontend/src/app.scss
@@ -133,7 +133,7 @@
 
   .kpm-muted-text {
     padding-top: 0.1em;
-    color: var(--kpmPaneTextMuted);
+    color: var(--kpmPaneTextSoft);
     font-size: 0.9em;
   }
 }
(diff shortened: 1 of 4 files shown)


kaj:
Note: It seems the prettier github action is out of date and don't really handle scss properly.  It works fine with my local prettier.

kpm/kpm-frontend/src/app.scss
```diff
@@ -133,7 +133,7 @@
 
   .kpm-muted-text {
     padding-top: 0.1em;
-    color: var(--kpmPaneTextMuted);
+    color: var(--kpmPaneTextSoft);
```
jhsware, line 136: Is there a reason for the name change? To me "muted" has more appropriate meaning than "soft".

jhsware (changes requested):
The core concept is outstanding. I have a question about a name change.

kpm/kpm-frontend/src/app.scss
kaj: It's not a naming change, it's a merge.  There was both a "muted" and a "soft" variant, and I think we should only have one kind of "text with less contrast".  As the gray level is closer to the one that was called soft i kept that name (I can change that if you think muted is the better name, but it won't make the diff smaler).

jhsware: Oh, yeah I realise I used it for different things, but that relates to our previous discussion where muted might have been a bit too light. The merge was a great idea! Personally I prefer "muted" over "soft", this is influenced by Bootstrap using `.text-muted` as a class. If you disagree, perhaps you could ask Carlos to chime in.

kaj: I feel strongly about the level of contrast, but don't really hava an opinion about the name, so I'll change it!  :-)

jhsware (approved):
LGTM

PR #151, opened 2023-03-13
Reset margin etc for image.

kaj:
intra (or specific page) adds margin to img, which breaks layout.

https://www.kth.se/om/work-at-kth/lediga-jobb/what:job/jobID:603744/where:4/


jhsware (approved):
LGTM

PR #159, opened 2023-04-11
Allow asset caching

jhsware:
Using parceljs namer plugin to force adding hash to bundle names.
Checks filenames in distProd and inject the generated ones in loader.
NOTE: it will pick the first on it finds. This works in prod because it is only bundled once, but needs improvement for development.


kaj (approved):
I'm still not sure just finding the first file with the correct suffix is the right thing to do.  But it may be good enough as a starting point.

Also, I don't see anything that sets a far expires header, but that may be as planned, get the names right first and commit to long-term caching after that?

PR #165, opened 2023-05-02
Default to kpm is on in `isMenuActive`.

kaj:
This should be merged not before #164, and released to prod as late as
possible but no later than the social release of 2023-05-03.


kpm/kpm-frontend/src/beta/App.tsx
```diff
@@ -25,9 +25,9 @@ async function doActivateMenu(
 
 function isMenuActive() {
   const cookies = document.cookie.split(/;\s+/);
-  return !!cookies.find(
-    (item) => item.startsWith("use_kpm=") && !item.endsWith("=")
-  );
+  // The cookie is 't' for active or 'f' for not active.
+  // the default in absense of cookie is _active_ from 2023-05-03.
+  return !cookies.find((item) => item.trim() == "use_kpm=f");
```
jhsware, line 30: - Prefer equality check with `===`
- Perhaps [.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is a better fit

jhsware (approved):
LGTM

PR #170, opened 2023-05-03
Dropdown menu size fix

jhsware:
This solves an issue where teachers with small laptop screens couldn't see the full Cours Admin dropdown menu.


kaj (approved):
LGTM!

PR #171, opened 2023-05-04
Check if session is still valid when computer is waking up from sleep

jhsware:
There is no event fired on wake up so this is a known work around.

The issue we are solving is when someone awakes their computer after a nights sleep with a page open containing a logged KPM. https://edge.sys.kth.se/itsm/EfecteFrameset.do#/workspace/datacard/view/173050210


kaj (approved):
Ser rimligt ut (efter förklaring) ...   Kanske siffran 2000 borde stoppas i en variabel, så vi inte missar att ändra både intervallet och ålderskollen om vi ändrar perioden?

jhsware:
Fixed in 864917

jhsware:
Let's close this, I don't think this is a good solution.

PR #172, opened 2023-05-08
Fix canvas rooms in studies.

kaj:
Actually filter the canvas rooms, rather than just showing whatever one happens to first in the json.

### DIFF
--- a/kpm/kpm-frontend/src/panes/studies.tsx
+++ b/kpm/kpm-frontend/src/panes/studies.tsx
@@ -134,6 +134,7 @@ function Course({ courseCode, course }: TCourseProps) {
   const status = course.completed ? "godkand" : roundToShow?.status;
   const exams = course.rooms?.filter((c) => c.type === "exam");
   const selTerm = `${roundToShow.year}${roundToShow.term}`;
+  // Show non-exams rooms for the selected term, but don't hide rooms w/o term.
   const filteredRooms = course.rooms?.filter(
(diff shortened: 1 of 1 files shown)


kpm/kpm-frontend/src/panes/studies.tsx
```diff
@@ -131,10 +131,12 @@ function RoundDesc({ round }: { round: TStudiesCourseRound }) {
 
 function Course({ courseCode, course }: TCourseProps) {
   const [roundToShow, setRoundToShow] = React.useState(course.rounds?.[0]);
-  // TODO: Try to find relevant room for selected round?
-  const roomToShow = course.rooms?.[0] || undefined;
   const status = course.completed ? "godkand" : roundToShow?.status;
   const exams = course.rooms?.filter((c) => c.type === "exam");
+  const selTerm = `${roundToShow.year}${roundToShow.term}`;
+  const filteredRooms = course.rooms?.filter(
+    (c) => c.type !== "exam" && (!c.startTerm || c.startTerm === selTerm)
```
jhsware, line 138: Perhaps add a note that !c.startTerm is used so we don't hide course rooms without terms (it isn't obvious unless one has had our discussion).

jhsware (changes requested):
Minor

kpm/kpm-frontend/src/panes/studies.tsx
kaj: Fixed.  Ok?

jhsware (approved):
LGTM

PR #175, opened 2023-05-12
We now get givenname in login claims

jhsware:
We now get givenname in login claims which allows us to show correct name in menu.

This solves KTH-INC-4307639


kaj (approved):
Yay!

PR #179, opened 2023-05-22
Use same data source for display name and menu options

jhsware:


kaj (commented):
Looks good!

Looking at remaining `...props` in app.tsx, the call to getRoutes, I think the _route_ for teaching and studies doesn't really need to be conditional, and numNewNotifications is not used anyway, so maybe the entire TRouterProps isn't needed?

I think the current conditionals in getRoutes may leed to a simliar situation to the current one, but instead of not rendering Studies or Teaching, they are rendered, but the route is missing to they can't be opened.  Do you agree that may be a risk?

If so, I think the prerequirements should be removed from routes.tsx lines 36 and 41, and after that I think that no actual fields of TRouterProps is still used, so it should be purged.  Right?

jhsware:
If we don't make them conditional, a user could navigate to those panes by typing in the adress bar. Do we want this?

PR #181, opened 2023-05-22
new pipeline trigger suggestion

dforslof:
För att inte trigga 4 pipeline jobb på en PR så önskar jag detta, ganska generellt så borde inte vara för strikt


kaj (commented):
I'm not sure ...  Seems reasonable at first.  But if there is a breaking change in one of the apis and we still use the old version in kpm, it won't be caught.  On the other hand, I don't think we catch those with the current setup either, do we?

PR #182, opened 2023-05-22
Use the actual ref URI to files

jhsware:


kpm/kpm-frontend/__test__/__snapshots__/test-panesProfile.spec.tsx.snap
```diff
@@ -49,11 +49,11 @@ exports[`<Profile /> Can be rendered 1`] = `
         <li>
           <a
             className="kpm-profile-link"
-            href="https://www.kth.se/profile/jhsware/"
+            href="https://www-r.referens.sys.kth.se/files/thumbnail/jhsware"
```
kaj, line 52: Oops!  This is wrong!

PR #189, opened 2023-05-26
Fix a missing xlat

kaj:
Fix missing translation for `Services for Students`.

Also remove a bunch of unused translations (for specific services, that comes pretranslated from the backend now).


jhsware (approved):
LGTM

PR #192, opened 2023-06-01
Sort services in columns.

kaj:
The main selected services list is two columns, not a bunch of rows, so "Annual ..." comes below "Administrate", not beside.  Also a bunch of cleanup of the main layout of the services panel (less calc, fewer properties, let the flex handling in the browser do the math).

https://trello.com/c/n0GW8mHA

### DIFF
--- a/kpm/kpm-frontend/src/panes/services.scss
+++ b/kpm/kpm-frontend/src/panes/services.scss
@@ -1,35 +1,18 @@
 @use "../vars" as *;
 
 #kpm-6cf53 div.kpm-services {
-  $rowGap: calc(2 * var(--kpmGap));
-  $colGap: calc(4 * var(--kpmGap));
-  $colCount: 3;
-  $colWidth: calc(100% / $colCount - $colGap * ($colCount - 1) / $colCount);
-
-  display: flex;
-  flex-flow: row wrap;
   align-items: flex-start;
-  gap: $rowGap $colGap;
 
   .kpm-col {
-    width: $colWidth;
-    flex-grow: 0;
+    flex: 13em 1;
   }
 
   .kpm-col.kpm-services-links {
-    width: calc(2 * $colWidth);
+    flex: 20em 6;
 
     & > ul {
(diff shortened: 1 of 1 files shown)


kpm/kpm-frontend/src/panes/services.scss
```diff
@@ -1,35 +1,18 @@
...
-      flex-flow: row wrap;
-      align-items: flex-start;
-      gap: $rowGap $colGap;
-
-      & > li {
-        $colCount: 2;
-        width: calc((100% - $colGap) / $colCount);
-        margin-bottom: 0;
-      }
+      column-width: 17em;
```
jhsware, line 14: The value feels picked at random. Would be great with a comment OR using a calculation from a variable that is used somewhere related.

kpm/kpm-frontend/src/panes/services.scss
```diff
@@ -1,35 +1,18 @@
...
-
-  display: flex;
-  flex-flow: row wrap;
   align-items: flex-start;
-  gap: $rowGap $colGap;
 
   .kpm-col {
-    width: $colWidth;
-    flex-grow: 0;
+    flex: 13em 1;
```
jhsware, line 7: This value also feels random. See other note.

jhsware (changes requested):
The result looks good, but I have some comments regarding understandability.

kpm/kpm-frontend/src/panes/services.scss
kaj: It is picked by trying different values and selecting one that looks good.  Not sure what to write in a comment about that, other than something like "this seems fine", which don't add much value ...

kpm/kpm-frontend/src/panes/services.scss
kaj: Same.  :-)

jhsware (approved):
LGTM

jhsware (approved):
LGTM

PR #195, opened 2023-06-30
Expose Canvas program rooms 

jhsware:
- refactored interfaces for improved readability
- added more typing to APIs
- exposing Canvas program rooms in API
- refactored Programme pane to show program rooms in Canvas

Do not merge and release until:
- program rooms are published
- students have been enrolled to program rooms

PR #199, opened 2023-07-20
Program rooms improved(?)

kaj:
This PR is my extended review and continuation of #195 .

Do not merge to master and release until:
* program rooms are published
* students have been enrolled to program rooms

This branch may be merged to #195 before that, though.

### DIFF
--- a/apis/my-canvas-rooms-api/src/interface.ts
+++ b/apis/my-canvas-rooms-api/src/interface.ts
@@ -15,6 +15,7 @@ export type APIUserErrType = "NotFound";
 export type APIUser = {
   courseRooms?: Record<string, Link[]>;
   programRooms?: Record<string, Link>;
+  otherRooms?: Link[];
 };
 
 /**
--- a/apis/my-canvas-rooms-api/src/interface.ts
+++ b/apis/my-canvas-rooms-api/src/interface.ts
@@ -13,8 +13,8 @@ export type APIMutedAuthErrType = "MissingAuthToken";
  */
 export type APIUserErrType = "NotFound";
 export type APIUser = {
-  courseRooms?: Record<string, Link[]> | null;
-  programRooms?: Record<string, Link> | null;
+  courseRooms?: Record<string, Link[]>;
+  programRooms?: Record<string, Link>;
 };
 
 /**
--- a/apis/my-canvas-rooms-api/src/interface.ts
+++ b/apis/my-canvas-rooms-api/src/interface.ts
@@ -13,9 +13,9 @@ export type APIMutedAuthErrType = "MissingAuthToken";
  */
 export type APIUserErrType = "NotFound";
 export type APIUser = {
-  courseRooms?: Record<string, Link[]>;
-  programRooms?: Record<string, Link>;
-  otherRooms?: Link[];
+  courseRooms: Record<string, Link[]>;
+  programRooms: Record<string, Link>;
+  otherRooms: Link[];
 };
 
 /**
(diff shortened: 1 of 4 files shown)


apis/my-canvas-rooms-api/src/interface.ts
```diff
@@ -13,8 +13,9 @@ export type APIMutedAuthErrType = "MissingAuthToken";
  */
 export type APIUserErrType = "NotFound";
 export type APIUser = {
-  courseRooms?: Record<string, Link[]> | null;
-  programRooms?: Record<string, Link> | null;
+  courseRooms?: Record<string, Link[]>;
+  programRooms?: Record<string, Link>;
```
kaj, line 17: Don't allow _both_ `undefined` and `null` value for the response fields.

apis/my-canvas-rooms-api/src/interface.ts
```diff
@@ -13,8 +13,9 @@ export type APIMutedAuthErrType = "MissingAuthToken";
...
+  otherRooms?: Link[];
```
kaj, line 18: ... and add `otherRooms` as a separate field for the fallback rooms, rather than treating them as course rooms for an unknown course.

kpm/kpm-frontend/src/panes/programme.tsx
```diff
@@ -100,15 +100,20 @@ type TProgramItemProps = { programCode: string; program: TCanvasRoom };
...
-      </a>
+      <p className="kpm-program-item-name">
+        <a
+          href={
+            typeof program.url === "string" ? program.url : program.url?.href
+          }
+        >
+          {displayName}
+        </a>
+      </p>
```
kaj, line 116: Make the name of the program room the link to the program room rather than putting the most important thing in a little link to the right.   I think this is so much of a UX benefit that it is worth a little inconsistency as to the course rooms (as there may be more than one room for each course but only one for each program, there will be some inconsistency anyway).

kpm/kpm-frontend/src/panes/programme.tsx
```diff
@@ -100,15 +100,20 @@ type TProgramItemProps = { programCode: string; program: TCanvasRoom };
...
       <a href={`https://www.kth.se/student/kurser/program/${programCode}`}>
         {i18n("Syllabus")}
```
kaj, line 118: Maybe repeat the program code here (e.g. "CDATE syllabus") to avoid having a bunch of same-word links?  On the other hand, many students will have only one syllabus anyway.

kaj (commented):
I wrote some motivation for most of my changes as a review here rather than on the original PR.

apis/my-canvas-rooms-api/src/interface.ts
jhsware: Lets discuss this.

apis/my-canvas-rooms-api/src/interface.ts
jhsware: Great idea!

kpm/kpm-frontend/src/panes/programme.tsx
jhsware: I agree that Clickable titles are preferable, but perhaps Carlos should way in on this change. @exacs

kpm/kpm-frontend/src/panes/programme.tsx
jhsware: I believe the idea is to be clear about where the link takes the user. A bit of repetition is okay IMO. As you noted, most users won't belong to a huge number of programs and they probably won't use this menu pane frequently enough to remember where the links take you. Keeping it as "Syllabus" for now is reasonable.

jhsware (changes requested):

jhsware (approved):
LGTM

PR #202, opened 2023-08-25
Use button for clickable area in filter

jhsware:


kaj (approved):
LGTM.  Maybe it would be a good idea to also replace the `nav ul` containing the filter to a `menu` ?

PR #205, opened 2023-08-31
Make course current earlier.

kaj:
Consider a course "current" from 14 (not just 5) dsys before the first tuition date.


jhsware (approved):
LGTM

PR #206, opened 2023-09-05
Feature show re-registrations

jhsware:
- show re-registrations in list of course rounds
- avoid duplicate round header with multiple rounds
- fix round header when no round iexists in course object

### DIFF
--- a/kpm/kpm-backend-interface/src/entities.ts
+++ b/kpm/kpm-backend-interface/src/entities.ts
@@ -20,7 +20,7 @@ export type TProgramCode = string;
 // QUESTION: Should we import types from the API-packages? Should these types be moved to separate packages?
 // Same as type Link in my-canvas-rooms-api/src/api.ts
 export type TCanvasRoom = {
-  url: string;
+  url: URL;
   name: string;
   state: "unpublished" | "available" | "completed" | "deleted";
   text?: string;
--- a/kpm/kpm-backend-interface/src/entities.ts
+++ b/kpm/kpm-backend-interface/src/entities.ts
@@ -20,7 +20,7 @@ export type TProgramCode = string;
 // QUESTION: Should we import types from the API-packages? Should these types be moved to separate packages?
 // Same as type Link in my-canvas-rooms-api/src/api.ts
 export type TCanvasRoom = {
-  url: URL;
+  url: string;
   name: string;
   state: "unpublished" | "available" | "completed" | "deleted";
   text?: string;
--- a/kpm/kpm-backend-interface/src/entities.ts
+++ b/kpm/kpm-backend-interface/src/entities.ts
@@ -53,17 +53,17 @@ export type TStudiesCourse = {
   completed: boolean;
   rounds: TStudiesCourseRound[]; // any rounds that are current comes first.
   rooms: TCanvasRoom[] | null;
+  current: boolean;
 };
 
 export type TStudiesCourseRound = {
-  status: "antagna" | "registrerade" | "omregistrerade";
+  status: "antagna" | "registrerade" | "omregistrerade" | "godkand";
   year: number;
   term: "1" | "2";
   ladokRoundId?: string;
   firstTuitionDate?: string; // "YYYY-MM-DD"
   lastTuitionDate?: string; // "YYYY-MM-DD"
   shortName?: string;
-  current: boolean;
 };
 
 export type TTeachingCourse = {
--- a/kpm/kpm-backend-interface/src/entities.ts
+++ b/kpm/kpm-backend-interface/src/entities.ts
@@ -56,7 +56,7 @@ export type TStudiesCourse = {
 };
 
 export type TStudiesCourseRound = {
-  status: "antagna" | "registrerade";
+  status: "antagna" | "registrerade" | "omregistrerade";
   year: number;
   term: "1" | "2";
   ladokRoundId?: string;
--- a/kpm/kpm-backend/src/api/studies.ts
+++ b/kpm/kpm-backend/src/api/studies.ts
@@ -21,11 +21,13 @@ const MY_STUDIES_API_URI =
   process.env.MY_STUDIES_API_URI || "http://localhost:3003/kpm/studies";
 const MY_STUDIES_API_TOKEN = process.env.MY_STUDIES_API_TOKEN!; // required by .env.in
 
+const COURSE_STATUS_TO_SHOW = ["registrerade", "omregistrerade"];
+
 // Copied from my-studies-api:
 export type TApiUserCourse = {
   type: "kurser";
   course_code: TCourseCode;
-  status?: "antagna" | "godkand" | "registrerade";
+  status?: "antagna" | "godkand" | "registrerade" | "omregistrerade";
   year?: number;
   term?: "1" | "2";
   round?: string;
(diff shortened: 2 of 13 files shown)


kpm/kpm-backend-interface/src/entities.ts
```diff
@@ -53,17 +53,17 @@ export type TStudiesCourse = {
   completed: boolean;
   rounds: TStudiesCourseRound[]; // any rounds that are current comes first.
   rooms: TCanvasRoom[] | null;
+  current: boolean;
 };
 
 export type TStudiesCourseRound = {
-  status: "antagna" | "registrerade" | "omregistrerade";
+  status: "antagna" | "registrerade" | "omregistrerade" | "godkand";
```
kaj, line 60: Godkänd är man på en kurs, så vad betyder godkänd här?  Att kursomgången på något sätt är relevant samtidigt som man är godkänd på kursen?

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -88,86 +89,57 @@ export async function studiesApiHandler(
...
-          continue;
-        }
-
-        // Skip if missing required fields
-        if (!role.status || !role.year || !role.term) {
-          continue;
-        }
+        // Skip entries that don't have a term or year, they aren't proper
+        // course rounds
+        if (role.year === undefined || role.term === undefined) continue;
```
kaj, line 101: Maybe clearer comment: "Skip roles that does not represent a course round".
Also note: every role where status is "godkand" will trigger this continue.

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -88,86 +89,57 @@ export async function studiesApiHandler(
...
-          role.round === undefined && role.status === "omregistrerade"
-            ? "term"
-            : role.round;
+        mytermrounds[term] ??= [];
 
-        // Skip if no round id can be determined
-        if (roleRound === undefined) {
-          continue;
-        }
+        // Find kopps round entry by round id hack...
```
kaj, line 107: Hack?  This is a proper lookup by the only id UG gives us.

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -88,86 +89,57 @@ export async function studiesApiHandler(
...
-        for (let [_id, round] of Object.entries(rounds)) {
-          if (round.current) {
-            current_rounds.push(round);
-          } else {
-            other_rounds.push(round);
-          }
+        const termRounds = reduceRoundsObject(rounds);
+
+        // If there is a round specific instance we ignore "omreg"
+        if (termRounds[2] || termRounds[1]) {
```
kaj, line 132: Is this supposed to mean "any proper courseround"?  If so, the roundid can be any single nonzero digit.

Maybe reduceRoundsObject should return a map of actual rounds _and_ an entirely separate flag for omreg, instead of putting a "fake" omreg round in the map?

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -88,86 +89,57 @@ export async function studiesApiHandler(
...
+          delete termRounds["omreg"];
         }
+
+        let tmp =
+          isRoundCurrent(termRounds[2]) || isRoundCurrent(termRounds[1]);
+        if (tmp === undefined) {
+          tmp = isRoundCurrent(termRounds["omreg"]);
+        }
```
kaj, line 140: With the above suggestion, this would be written as:

```
  if (len(termRounds) > 0) {
    courseIsCurrent ||= termRounds.any(isRoundCurrent);
  } else {
    courseIsCurrent ||= isTermCurrent(term);
  };
```

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -191,7 +164,66 @@ export async function studiesApiHandler(
...
+          lastTuitionDate: getTermEndDate(year, term),
+          shortName: "omreg_lbl",
+        };
+        break;
+    }
+  }
+  return termRounds;
+}
+
+function isRoundCurrent(inp: any) {
```
kaj, line 222: With the above suggested change, isRoundCurrent could be properly typed.   Or we could write a new `TuitionPeriod` type that would match both a round and a term.

kaj (commented):
This is an improvement!  Still some things to discuss, though ...

kpm/kpm-backend/src/api/studies.ts
jhsware: FIXED in 14a9d2e

kpm/kpm-backend/src/api/studies.ts
jhsware: Each entry has a property `ladokUID` which I believe would be better to use as an identifier, but we don't get that ID from the UG from what I can see.

kpm/kpm-backend/src/api/studies.ts
jhsware: It is a term so perhaps the naming of the variable is confusing. Let's change to a better name!

kpm/kpm-backend/src/api/studies.ts
jhsware: `termRounds` is an object so my thoughts were that this "dumb" solution will be easier to read.

kpm/kpm-backend/src/api/studies.ts
jhsware: I was knee deep in fixing type issues so I cheated.

kpm/kpm-backend-interface/src/entities.ts
jhsware: Löste ett typningsproblem och att mutera hela skiten kändes risky.

kpm/kpm-backend/src/api/studies.ts
jhsware: FIXED in 879658a

kpm/kpm-backend/src/api/studies.ts
jhsware: Changed in 879658a

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -159,12 +152,80 @@ export async function studiesApiHandler(
...
+      firstTuitionDate,
+      lastTuitionDate,
+      shortName,
+    } = round;
+
+    if (status === undefined) continue;
+    if (year === undefined) continue;
+    if (term === undefined) continue;
+
+    switch (status) {
```
kaj, line 179: I don't see the `"antagna"` case here?  I think it should be?  Or why are cases handled so different dependeing on status anyway?  The only thing is think shoud be done about `status` is to select the "most relevant" if there are multiple statuses for the same course round.  So the `termRounds[ladokRoundId] = { ... }` should be done in all cases where there is a `ladokRoundId`, except the status should be prio-selected among any previous value and the new value.

kaj (commented):
Just two things before the retro.  We'll continue later ...

kpm/kpm-backend/src/api/studies.ts
```diff
@@ -86,58 +89,47 @@ export async function studiesApiHandler(
...
       for (let [_term, rounds] of Object.entries(mytermrounds)) {
-        for (let [_id, round] of Object.entries(rounds)) {
-          if (round.current) {
-            current_rounds.push(round);
-          } else {
-            other_rounds.push(round);
-          }
-        }
+        const termRounds = reduceRoundsObject(rounds);
+        courseIsCurrent = termRounds.some((o) => isRoundCurrent(o));
```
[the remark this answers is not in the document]
jhsware: FIXED in 2a792c2

kpm/kpm-backend/src/api/studies.ts
jhsware: FIXED in 9676d7a

kaj (approved):
Ser typ rimligt ut ...  Men det har blivit ganska långt ...

kpm/kpm-frontend/src/panes/studies.tsx
```diff
@@ -120,49 +114,69 @@ type TCourseProps = {
...
-function RoundDesc({ round }: { round: TStudiesCourseRound }) {
+function RoundDesc({ round }: { round?: TStudiesCourseRound }) {
+  if (round === undefined)
+    return <React.Fragment>{i18n("om_kursen")}</React.Fragment>;
+
   return (
     <React.Fragment>
       {i18n("term" + round.term)}
-      {round.year % 100} ({round.shortName || round.ladokRoundId})
+      {round.year % 100} ({i18n(round.shortName) || round.ladokRoundId})
```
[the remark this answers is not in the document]
jhsware: FIXED in 0b2019941

PR #209, opened 2023-09-19
Trap focus within menu pane when tabbing

jhsware:
This only supports tabbing. We don't support:
- key nav in menu pane
- ~~esc to close~~ (FIXED in [c086b5c](https://github.com/KTH/kpm/pull/209/commits/c086b5c47fd438eb9d6f6cd838f808fb91e00ee0))

### DIFF
--- a/kpm/kpm-frontend/__test__/__snapshots__/test-panesGroups.spec.tsx.snap
+++ b/kpm/kpm-frontend/__test__/__snapshots__/test-panesGroups.spec.tsx.snap
@@ -2,6 +2,7 @@
 
 exports[`<Groups /> Can be rendered 1`] = `
 <dialog
+  aria-modal="true"
   className="kpm-modal"
 >
   <button
@@ -91,6 +92,7 @@ exports[`<Groups /> Can be rendered 1`] = `
 
 exports[`<Groups /> Can be rendered with data and none starred 1`] = `
 <dialog
+  aria-modal="true"
   className="kpm-modal active"
 >
   <button
@@ -220,6 +222,7 @@ exports[`<Groups /> Can be rendered with data and none starred 1`] = `
 
 exports[`<Groups /> Can be rendered with data and one starred 1`] = `
 <dialog
+  aria-modal="true"
   className="kpm-modal active"
 >
   <button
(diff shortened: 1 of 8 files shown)


kaj (commented):
Hm.  Känns lite oroande att implementera tab-to-focus i eget javascript och stänga av browserns hantering i stället för att fixa browserns hantering.

Och hur relaterar det här till förslaget att sätta in den öppna panelen "rätt i ordningen" mellan det valda toplevelalternativet och nästa, som @exacs har föreslagit?

exacs:
I'd say this cannot be released to production until it is possible to close the modal with <kbd>ESC</kbd> key OR a "close button" is actionable inside the modal.

In other words: According to WCAG, users **should** be able to close modals with keyboard-only (either providing a close button, or by pressing ESC, or both).

jhsware:
ESC to close supported with [c086b5c](https://github.com/KTH/kpm/pull/209/commits/c086b5c47fd438eb9d6f6cd838f808fb91e00ee0)

@exacs @kaj when we close the menu pane with ESC, we don't set the focus to the menu item. A user needs to tab from the start of the focus loop. This behaviour is also found in dropdowns, if you have tabbed into the dropdown and press ESC you loose focus. I get that this isn't optimal, but is it acceptable for now?

exacs:
Hmm... Sorry in advance if I sound grumpy!, but [this page in WCAG](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F85#:~:text=Failure%20Example%202,menu%20was%20opened.), describes what is considered to be a failure in the SC 1.4.3 Focus Order:

TL;DR: it is considered a failure if, after dismissing a dialog, the control does not return to the button that opened it _or_ the control does not go to that button after pressing <kbd>Shift + Tab</kbd> (tab back) few times

jhsware:
@exacs I actually appreciate that we are being ambitious on this!

jhsware:
FIXED restoring focus to menu item when closing in 5eeed203f and [727e95f](https://github.com/KTH/kpm/pull/209/commits/727e95f752485d1329c5848a310b71778f630748)

PR #210, opened 2023-09-19
Fix mobile root menu behaviour

jhsware:
Keep the root menu open when navigating to child menu panes.


jhsware:
We need to check the router state somehow, the hosting page might set the hash and in which case the root menu will always be open by default.

jhsware:
Merging because I believe all issues have been adressed and further improvements can be made as separate PRs.

PR #217, opened 2023-11-21
Refactor/type safe fetcher

exacs:


exacs:
- `loading` boolean is kept for retro-compatibility reasons, may disappear
- not sure if it is "correct" to use a hook _after_ a `return`. I need to test what happens

PR #218, opened 2023-11-21
Restrict current studies to antagna

exacs:
This PR adds extra logic to limit the amount of courses considered "current"


jhsware (approved):
LGTM

PR #225, opened 2024-04-16
Remove beta

exacs:
This PR removes the possibility to switch to the "old" personal menu.

It also removes the interrobang icon


kaj (approved):
Looks good to me.  We should also update the text on the info page, that can be done either in this pr or a separate one.  (And yes, there should still _be_ an info page, even if the prominent link to it is removed).

PR #227, opened 2024-04-24
Read current site without using RegEx

exacs:
This PR changes the way kpm guesses which is the current "site" (external, student-web or intranet).

Instead of looking at the URL, it looks if a given class is present in the header.

### DIFF
--- a/kpm/kpm-backend/src/widget.js/notLoggedIn.js
+++ b/kpm/kpm-backend/src/widget.js/notLoggedIn.js
@@ -1,25 +1,26 @@
 (function (js, css, url) {
-  let RS = /^https?:\/\/(www\.)?kth.se\/(en\/)?student($|\/)/;
-  let RE = /^https?:\/\/(www\.)?kth.se($|\/)/;
-  let RI = /^https?:\/\/intra\.kth\.se/;
   let lang = document.documentElement.lang.startsWith("sv") ? "sv" : "en";
-  let links = {
-    external: {
+  let links = [
+    {
+      className: "external",
       href: { sv: "https://www.kth.se", en: "https://www.kth.se/en" },
       label: { en: "kth.se", sv: "kth.se" },
     },
-    student: {
+    {
+      className: "student-web",
       href: {
         sv: "https://www.kth.se/student",
         en: "https://www.kth.se/en/student",
       },
       label: { en: "Student web", sv: "Studentwebben" },
     },
-    intra: {
+    {
+      className: "intranet",
       href: { sv: "https://intra.kth.se", en: "https://intra.kth.se/en" },
       label: { en: "Intranet", sv: "Intranät" },
     },
-  };
+  ];
+
   let lbls = {
     w: { en: "Websites", sv: "Webbplatser" },
     l: { en: "Login", sv: "Logga in" },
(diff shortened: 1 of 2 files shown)


kaj:
How about canvas and other apps that don't (yet) have the new kthstyle header?  I guess they will just be "none of the above", and that is probably fine?

(I just want that to be a concious decision, rather than just happening)

exacs:
@kaj Yes, Canvas and sites without any of the headers will be "none". All links will be "default" (blue text, transparent background), none of them marked.

kpm/kpm-frontend/src/components/entrances.tsx
```diff
@@ -14,15 +10,11 @@ const labels = {
...
-  if (STUDENT_REGEX2.test(currentPage)) {
-    return "student";
-  } else if (INTRA_REGEX.test(currentPage)) {
-    return "intra";
-  } else if (EXTERNAL_REGEX.test(currentPage)) {
-    return "external";
-  }
+  return SITES.find((site) =>
+    document.querySelector("kth-header")?.classList.contains(site)
+  );
```
jhsware, line 17: Do we have a custom element called "kth-header" or is this query selector a typo?

kpm/kpm-frontend/src/components/entrances.tsx
```diff
@@ -14,15 +10,11 @@ const labels = {
 
 /** Get the current site */
 function getCurrentSite(): "external" | "intra" | "student" | undefined {
-  const currentPage = window.location.toString();
+  const SITES = ["external", "intra", "student"] as const;
```
jhsware, line 13: The site names don't match the classnames in the variable links. Considering the check classList.contains(site) this feels strange that they are so close but still not the same. If it is intentional, perhaps one could add a comment.

kpm/kpm-frontend/src/components/entrances.tsx
jhsware: notLoggedIn.js:L52 indicates typo.

kpm/kpm-backend/src/widget.js/notLoggedIn.js
```diff
@@ -48,40 +49,29 @@
...
     .join("");
 
   let login = `<a class="kth-menu-item kpm-login" href="${url}?nextUrl=${location.href}">${lbls.l[lang]}</a>`;
-
-  let currentLabel =
-    current === "" ? links.external.label[lang] : links[current].label[lang];
+  let currentItem = links.find((link) =>
+    header?.classList.contains(link.className)
+  );
+  let currentLabel = currentItem?.label[lang] ?? links[0].label[lang];
```
jhsware, line 74: Should we have a fallback?

jhsware (changes requested):
I have some minor questions about the code. See comments.

kpm/kpm-frontend/src/components/entrances.tsx
```diff
@@ -4,16 +4,16 @@ import "./entrances.scss";
...
-  student: i18n("shortcut.studentweb"),
-};
+  intranet: i18n("shortcut.intranet"),
+  "student-web": i18n("shortcut.studentweb"),
+} as const;
 
 /** Get the current site */
-function getCurrentSite(): "external" | "intra" | "student" | undefined {
-  const SITES = ["external", "intra", "student"] as const;
+function getCurrentSite(): keyof typeof labels | undefined {
```
jhsware, line 12: Snyggt!

jhsware (approved):
LGTM

NOTE: I haven't test run the code.

PR #229, opened 2024-05-31
Fix: accessible mobile menu

exacs:
This PR fixes these accessibility problems:

- The button to open the personal menu in mobile was not focusable.

  `<a>` tags without `href` are not focusable. This PR replaces the element with `<button>`

- The links for personal menu (Studies, Teaching...) were focusable even when they were not visible

  The links were "non-visible" by making the container shorter than the content, which meant that buttons were still focusable. This PR adds `display: none` to the container

### DIFF
--- a/kpm/kpm-frontend/src/Menu.scss
+++ b/kpm/kpm-frontend/src/Menu.scss
@@ -42,7 +42,7 @@ $icon-kpm-notifications: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%
   box-shadow: 0 0.25rem 0.25rem 0 rgb(0 0 0 / 20%);
 }
 
-#kpm-6cf53 .kpm-mobile-menu {
+#kpm-6cf53 .kpm-mobile-menu-opener {
   display: none;
 }
 
@@ -116,7 +116,6 @@ $icon-kpm-notifications: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%
 }
 
 #kpm-6cf53 {
-  .kpm-mobile-menu.kpm-mobile,
   .kpm-profile-item.kpm-mobile,
   .kpm-mobile-logout.kpm-mobile {
     display: none;
@@ -136,7 +135,7 @@ $icon-kpm-notifications: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%
     display: none !important;
   }
 
-  #kpm-6cf53 .kpm-mobile-menu {
+  #kpm-6cf53 .kpm-mobile-menu-opener {
     display: flex;
     justify-content: flex-end;
   }
--- a/kpm/kpm-frontend/src/Menu.scss
+++ b/kpm/kpm-frontend/src/Menu.scss
@@ -150,12 +150,15 @@ $icon-kpm-notifications: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%
     }
 
     &.active > ul {
-      display: block;
+      visibility: visible;
+      opacity: 1;
     }
   }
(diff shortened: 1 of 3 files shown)


kpm/kpm-frontend/src/Menu.tsx
```diff
@@ -74,27 +74,27 @@ export function Menu() {
...
+                  src={prefixHost(
+                    "www",
+                    `/files/thumbnail/${currentUser?.username}`
+                  )}
+                  alt=""
+                  className="kpm-profile-image"
+                />
+                <span>{i18n("Personal menu")}</span>
+              </button>
+            </div>
```
jhsware, line 95: Since we are moving the button out of the list (which feels reasonable), perhaps we should use \<header\> for this container instead of a generic div to emphasize the relationship with the list.

kpm/kpm-frontend/src/Menu.scss
```diff
@@ -140,16 +148,16 @@ $icon-kpm-notifications: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%
...
       height: 100vh;
     }
+
+    &.active > ul {
+      display: block;
+    }
   }
 
   body #kpm-6cf53 .kpm-menu > ul {
+    display: none;
```
jhsware, line 158: If we use `visibility: hidden` instead of `display: none`, we provide an opportunity to add a subtle animation when opening the menu:
```
visibility: hidden;
opacity: 0;
transition: opacity .3s;
```
and the corresponding:
```
visibility: visible;
opacity: 1;
```
on active.

The benefit of hiding the ul from layout calculations with `display: none` is not very high in this case.

jhsware (changes requested):
In general I think this is a great improvement. I have tested the interactions in the browser and they work as expected.

I have added a suggestion and a comment. Neither require any changes to approve this PR so once you have decided how you feel about the comments you can merge this without further review. Marking this with "request changes" is only so you don't miss the comments :)

kpm/kpm-frontend/src/Menu.tsx
exacs: I disagree with the use of `<header>` in this case for those reasons:

1. Landmarks (like `header`, `nav`, `main`, `footer`, etc.) does not give any semantic meaning when used inside another landmark (in this case it would be `header` inside `nav`). So, it is equivalent as `div`
2. Landmarks should be used to mark **important areas in a page**. Screen readers can show a list of all landmarks in a page, and showing too many of them it is actually bad.

To mark the relationship of the button and the list, in this case, it is better to use [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls)

kpm/kpm-frontend/src/Menu.scss
exacs: That can work! Btw, check these new CSS things that are interesting regarding animating `display`:

- [`transition-behavior`](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-behavior)
- [`@starting-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style)

Those are available in Chrome and Safari already and now in Firefox Nightly!

You can read [a better explanation in this article](https://nerdy.dev/using-starting-style-and-transition-behavior-for-enter-and-exit-stage-effects)

exacs:
I will merge this PR even when the "KPM ref" pipeline did not succeed. It seems that the security check is failing due to outdated dependencies. Will fix dependencies in a different branch or main

kpm/kpm-frontend/src/Menu.tsx
jhsware: Excellent!

PR #232, opened 2024-06-17
Prevent kpm to do anything when loaded inside <iframe>

exacs:
This PR checks if the current document is an iframe in the beginning of all. That means no React, no CSS is loaded in that case


jhsware (approved):
LGTM

PR #235, opened 2024-06-19
Fix/dropdown

exacs:
This PR fixes two problems with dropdowns:

- Dropdowns not visible when the modal is too small and the content of the dropdown is too big
- Too little padding around dropdowns when they are too big

### DIFF
--- a/kpm/kpm-frontend/src/components/groups.scss
+++ b/kpm/kpm-frontend/src/components/groups.scss
@@ -34,7 +34,10 @@
     z-index: 999;
     max-width: 95%;
     min-height: 200px; // Use a constant value in pixels, otherwise we have problems
-    max-height: calc(100vh - 20px); // Same as PADDING in groupsUtils.ts
+
+    // Same as PADDING in groupsUtils.ts
+    $padding-in-group-utils: 60px;
+    max-height: min(calc(90vh - $padding-in-group-utils), 40rem);
     background-color: var(--kpmPaneBg);
     overflow-y: auto;
     box-shadow: 4px 4px 10px rgb(0 0 0 / 20%);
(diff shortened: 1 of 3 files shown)


kpm/kpm-frontend/src/components/groups.scss
```diff
@@ -34,7 +34,10 @@
     z-index: 999;
     max-width: 95%;
     min-height: 200px; // Use a constant value in pixels, otherwise we have problems
-    max-height: calc(100vh - 20px); // Same as PADDING in groupsUtils.ts
+
+    // Same as PADDING in groupsUtils.ts
+    $padding-in-group-utils: 60px;
+    max-height: min(calc(90vh - $padding-in-group-utils), 40rem);
```
jhsware, line 40: Won't this set a maximum height of 40rem? Shouldn't it really be max(...)?

"The min() CSS function lets you set the smallest (most negative) value from a list of comma-separated expressions as the value of a CSS property value."
https://developer.mozilla.org/en-US/docs/Web/CSS/min

jhsware (changes requested):
Double check if you really intended to use min(...). Other than that it looks good to me.

kpm/kpm-frontend/src/components/groups.scss
exacs: My intention was to add a "maximum" value for the height, in a way that the dropdown:

- Is always fully visible (therefore less than 90% of the height)
- Is not too large: (less than `40rem`)

Otherwise: you have to scroll the container to see the whole dropdown, and you need to scroll to the bottom because the dropdown repositions itself

jhsware: Makes sense! Mission accomplished

jhsware (approved):
🚀

